PR #1, opened 2018-08-22
Import meta require

MylesBorins:
From https://github.com/nodejs/node/pull/21317

> Hey All,
> 
> This builds on top of @targos's earlier implementation of `import.meta.require`. It has been refactored to work with our current layout for the module system. I've also added some basic tests, let me know if you think we need something more robust.
> 
> @nodejs/modules I am not planning to land this without consensus from the group, but I think that this is a necessary base feature no matter what implementation we move forward with. I'll open a corresponding issue to discuss.

### DIFF
--- a/doc/api/esm.md
+++ b/doc/api/esm.md
@@ -45,6 +45,9 @@ The `import.meta` metaproperty is an `Object` that contains the following
 property:
 
 * `url` {string} The absolute `file:` URL of the module.
+* `require` {Function} To require CommonJS modules. This function enables
+  interoperability between CJS and ESM. See [`require()`]. None of the properties
+  generally exposed on require are available via `import.meta.require`.
 
 ### Unsupported
 
@@ -256,3 +259,4 @@ in the import tree.
 [Node.js EP for ES Modules]: https://github.com/nodejs/node-eps/blob/master/002-es-modules.md
 [addons]: addons.html
 [dynamic instantiate hook]: #esm_dynamic_instantiate_hook
+[`require()`]: modules.html#modules_require
--- a/doc/api/esm.md
+++ b/doc/api/esm.md
@@ -46,8 +46,9 @@ property:
 
 * `url` {string} The absolute `file:` URL of the module.
 * `require` {Function} To require CommonJS modules. This function enables
-  interoperability between CJS and ESM. See [`require()`]. None of the properties
-  generally exposed on require are available via `import.meta.require`.
+  interoperability between CJS and ESM. See [`require()`]. None of the
+  properties generally exposed on require are available via
+  `import.meta.require`.
 
 ### Unsupported
 
--- a/lib/internal/process/esm_loader.js
+++ b/lib/internal/process/esm_loader.js
@@ -1,12 +1,16 @@
 'use strict';
 
+const { makeRequireFunction } = require('internal/modules/cjs/helpers');
+const Module = require('module');
 const { internalBinding } = require('internal/bootstrap/loaders');
+
 const {
   setImportModuleDynamicallyCallback,
   setInitializeImportMetaObjectCallback
 } = internalBinding('module_wrap');
+const { defineProperty } = Object;
 
-const { getURLFromFilePath } = require('internal/url');
+const { getURLFromFilePath, getPathFromURL } = require('internal/url');
 const Loader = require('internal/modules/esm/loader');
 const path = require('path');
 const { URL } = require('url');
@@ -27,6 +31,27 @@ function initializeImportMetaObject(wrap, meta) {
   if (vmModule === undefined) {
     // This ModuleWrap belongs to the Loader.
     meta.url = wrap.url;
+    let req;
+    defineProperty(meta, 'require', {
+      enumerable: true,
+      configurable: true,
+      get() {
+        if (req !== undefined)
+          return req;
+        const url = new URL(meta.url);
+        const path = getPathFromURL(url);
+        const mod = new Module(path, null);
+        mod.filename = path;
+        req = makeRequireFunction(mod).bind(null);
+
+        // remove properties that don't affect ESM loading
+        // delete req.cache;
+        // delete req.extensions;
+        // delete req.main;
+        // delete req.resolve.paths;
+        return req;
+      }
+    });
   } else {
     const initializeImportMeta = initImportMetaMap.get(vmModule);
     if (initializeImportMeta !== undefined) {
--- a/lib/internal/process/esm_loader.js
+++ b/lib/internal/process/esm_loader.js
@@ -42,13 +42,8 @@ function initializeImportMetaObject(wrap, meta) {
         const path = getPathFromURL(url);
         const mod = new Module(path, null);
         mod.filename = path;
+        mod.paths = Module._nodeModulePaths(path);
         req = makeRequireFunction(mod).bind(null);
-
-        // remove properties that don't affect ESM loading
-        // delete req.cache;
-        // delete req.extensions;
-        // delete req.main;
-        // delete req.resolve.paths;
         return req;
       }
     });
--- a/test/es-module/test-esm-import-meta.mjs
+++ b/test/es-module/test-esm-import-meta.mjs
@@ -3,20 +3,47 @@
 import '../common';
 import assert from 'assert';
 
+const fixtures = import.meta.require('../common/fixtures');
+
 assert.strictEqual(Object.getPrototypeOf(import.meta), null);
 
-const keys = ['url'];
+const keys = ['url', 'require'];
 assert.deepStrictEqual(Reflect.ownKeys(import.meta), keys);
 
 const descriptors = Object.getOwnPropertyDescriptors(import.meta);
+
 for (const descriptor of Object.values(descriptors)) {
   delete descriptor.value; // Values are verified below.
-  assert.deepStrictEqual(descriptor, {
-    enumerable: true,
-    writable: true,
-    configurable: true
-  });
 }
+
+assert.deepStrictEqual(descriptors.url, {
+  enumerable: true,
+  writable: true,
+  configurable: true
+});
+
+assert.deepStrictEqual(descriptors.require, {
+  get: descriptors.require.get,
+  set: undefined,
+  enumerable: true,
+  configurable: true
+});
+
+assert.strictEqual(import.meta.require.cache, undefined);
+assert.strictEqual(import.meta.require.extensions, undefined);
+assert.strictEqual(import.meta.require.main, undefined);
+assert.strictEqual(import.meta.require.paths, undefined);
 
 const urlReg = /^file:\/\/\/.*\/test\/es-module\/test-esm-import-meta\.mjs$/;
 assert(import.meta.url.match(urlReg));
+
+const a = import.meta.require(
+  fixtures.path('module-require', 'relative', 'dot.js')
+);
+const b = import.meta.require(
+  fixtures.path('module-require', 'relative', 'dot-slash.js')
+);
+
+assert.strictEqual(a.value, 42);
+// require(".") should resolve like require("./")
+assert.strictEqual(a, b);


zenparsing:
Thanks for the example @michael-ciniawsky , I was wondering what it would look like.

What's the value in making users perform this `createRequireFunction` incantation in order to do something that they will need to do for interoperability purposes? `import.meta.require` is searchable and obviously not a web thing.

Generally speaking, if we provide a feature because users need it, then we shouldn't also punish them for using it by intentionally making it difficult to use.

michael-ciniawsky:
It's definitely debatable and if users absolute dislike (e.g shown by a survey) it, it's definitely better to favor DX and UX over long-term concerns in regard to potential technical debt etc. Still `createRequire` should be put on the agenda before this lands as it is an considerable alternative imho. Ideally with a vote by `nodejs/modules` to choose between one of the two for the minimal kernel implementation afterwards (to avoid getting lost in endless bikeshedding 😛)

devsnek:
i'd like to make sure we have this repo set up to automatically pull updates from nodejs/node before we're merging changes

bmeck:
@zenparsing This is not just about UX/DX. `createRequireFunction` has utility that `import.meta.require` is unable to replicate:

### 1. Web Compatibility Path:

`createRequireFunction` is able to allow developers to write ESM code that can function in both client and server environments without modification to the ESM source text:

```mjs
import {createRequireFunction} from 'module';
createRequireFunction(import.meta.url);
```

```mjs
// web 'module'
let base = import.meta.url;
let bundle = {
  '/foo.js': ...
}
export function createRequireFunction(url) {
  return function require(path) {
    return bundle[resolve(base, path)];
  };
}
```

`import.meta.require` is unable to do likewise because you cannot create a shim of any kind for it by its placement on `import.meta`. The only way to achieve this is to require source code transformations of ESM on the web.

### 2. Static analysis

`import.meta.require` has the same level of static analyzability as `import.meta.require` at a naive level. However, module namespaces are not mutable and thus do not have to deal with an entire category of static analysis bail outs that `import.meta.require` must deal with:

```mjs
// don't bail out here
import.meta.require('fs');
import.meta.require = function () {};
// bail out here
import.meta.require('fs');


// using window.require
require('fs');

// import.meta.require
var require = import.meta.require;
```

This is only a realistic compatibility concern if people choose to have incompatible implementations of `import.meta.require` due to using things like AMD, etc. However, the alternative doesn't suffer this complexity.

### 3. Simpler code mods / upgrade path

Tools and/or humans can add a simple prefix to files if they wish to use `require` in the expected ways without facing bailout concerns:

```mjs
import {createRequireFunction} from 'module';
var require = createRequireFunction(import.meta.require);
```

Doing so with `import.meta.require` faces the bailout complexity mentioned previously.

### 4. Better Errors

`import.meta.require` throws errors at runtime when it is not defined. This allows for code to be written in such a way that problematic usage is not found until after an application has fully started up, such as in the case of lazy requires:

```mjs
db.onconnect = (conn) => {
  // errors here
  const models = import.meta.require('models')(conn);
};
```

The more conventional way of using `createRequireFunction` would be to use static `import`:

```mjs
// errors before any evaluation
import {createRequireFunction} from 'module';
var require = createRequireFunction(import.meta.require);
db.onconnect = (conn) => {
  const models = require('models')(conn);
}
```

It should be noted that `createRequireFunction` faces similar problems at runtime if you do  dynamic import, but I don't know of a good reason to use dynamic import for this:

```mjs
// I don't know of any reason to do this?
// maybe just to see if `module` exists???
// but if you need to use `require` what help is that
var {createRequireFunction} = await import('module');
```

### 5. More complete feature detection

Even though it is statically analyzable, sometime feature detection is desirable for modules. This feature is still preserved using module namespaces.

```mjs
import * as module from 'module';
// note that this is able to be analyzed even if `module` is a shim
// this is because we can statically analyze what `module` exports
if (module.createRequireFunction) {
  // ...
}
```

`import.meta.require` is unable to be shimmed and thus faces problems with feature detection if you are unable to determine the deployment environment.

----

In addition, even with the added utility, `createRequireFunction` has a better upgrade story towards isomorphic code. We should aim to encourage users to write compatible code. This is not by any means an attempt at "punishment".

To disallow code that uses `createRequireFunction` you can use a loader that changes the module record resolved for `"module"` to be one without `createRequireFunction`, one that has deprecation warning, one that uses a shim, etc.. In order to disable code that uses `import.meta.require` you do not have similar approaches, and instead must rely on the rudimentary `undefined is not a function` error that users know well. The only alternative is to add a prefix to the source text being loaded during loading:

```mjs
{
  const _require = import.meta.require;
  // wrap it or replace it:
  import.meta.require = wrap(_require);
}
```

However, these once again reach bailout considerations mentioned above.

----

Overall, we should encourage simple usage of code that migrates towards compatibility stories in multiple environments and tooling. However, I do not believe that `import.meta.require` works towards that goal. `createRequireFunction` has simplistic usage if you do the prefix behavior mentioned above and does not require rewriting all usages of `require` to be `import.meta.require` to prevent bailout possibilities.

mcollina:
@bmeck can we call it module.createRequire(url) instead of createRequireFunction?

(+1 on the approach anyway)

WebReflection:
The usage of synchronous XHR is discourage and the creation of that `module` file would need bundlers regardless, meaning that **Web Compatibility** is a myth here.

No developer ever used `require` right away if not for development purposes and these days the majority of developers use bundlers.

The **static analysis** compares apples to oranges and gives for granted every developer will implement properly that `require` function.

The moment you have to tell developers how to properly create such function is the moment @zenparsing is right.

> Generally speaking, if we provide a feature because users need it, then we shouldn't also punish them for using it by intentionally making it difficult to use.

The `import.meta.require` **is** the simpler code mods / upgrade path, and there are no magic better errors handling in user land + feature detection is another myth because the word import cannot be used at all in non ES6+ code but `import.meta.require` can be easily found, replaced, patched both via bundlers or synchronous loaders.

TL;DR all @bmeck points are **very** subjective and some point does not solve any real-world use case while `import.meta.require` would.

mcollina:
I've quite a few usage of `require('module').createRequire(url)` or even better `require('module').createRequire(path)` that are not ESM specific. This can help with unit testing and enable frameworks to create better isolation right now. Such a function enables a Node.js script to "see" what a script located in another folder could require. This enables a level of flexibility that is currently not possible or very hard to achieve.

Consider this code (full example at https://github.com/fastify/fastify/blob/master/test/internals/handleRequest.test.js):

```js
const t = require('tap')
const test = t.test
const internals = require('../../lib/handleRequest')[Symbol.for('internals')]
const Request = require('../../lib/request')
const Reply = require('../../lib/reply')
const buildSchema = require('../../lib/validation').build
const Schemas = require('../../lib/schemas')
```

Would it be better if rewritten as:

```js
const t = require('tap')
const test = t.test
const project = require('module').createRequire(new URL(`file://${__dirname}/../../`)

// this block can now be copy-pasted in various part of the tests, and it is independent of
// the location of the file itself
const internals = project('./lib/handleRequest')[Symbol.for('internals')]
const Request = project('./lib/request')
const Reply = project('./lib/reply')
const buildSchema = project('./lib/validation').build
const Schemas = project('./lib/schemas')
```

`import.meta.require` _might_ be the way to go, but it's a more specific mechanism and one does not prohibit the other. I would love to have `require('module').createRequire()` in core right now.

MylesBorins:
I agree with Matteo, the two fearures are not mutually exclusive.

I'd personally like to see makeRequire land as part of minimal kernel (potentially land on core soon). We can then dig into specifics of import meta require as part of a larger vision... We could also implement import.meta.require with makeRequire 😂

WebReflection:
@mcollina this

```js
const project = require('module').createRequire(new URL(`file://${__dirname}/../../`)
```

is basically

```js
const project = source => require(`file://${__dirname}/../../${source}`);
```

which is even shorter to write.

That doesn't look like a compelling reason to block shipment of `import.meta.require`.

I am also not saying there are no use cases whatsoever for a `createRequire`, I am saying `import.meta.require` is what developers need/like and a better/easier way to move forward, IMO.

mcollina:
@WebReflection `require('module').createRequire()` is aware of `node_modules`, while your example is not.

```js
const project = require('module').createRequire(new URL(`file://${__dirname}/../../`)
const myLibrary = project('awesome-library') // is resolved from `file://${__dirname}/../../`
```

This could be _very_ helpful in a lot of cases (CLI tools etc).

A similar capability is available in the ecosystem through https://github.com/sindresorhus/resolve-from which is downloaded 6 millions time per month, and it's currently based on _undocumented functions_: https://github.com/sindresorhus/resolve-from/blob/60cd04e69135b96b98b848fff719b1276a5610c0/index.js#L29. 

IMHO we should just add this feature to core, as it will have the side benefit of reducing the usage of private functionality. I'd kindly ask @bmeck to send a PR to core for that if he is willing.

> I am also not saying there are no use cases whatsoever for a createRequire, I am saying import.meta.require is what developers need/like and a better/easier way to move forward, IMO.

Have you got any evidence of that? How can we collect evidence?

WebReflection:
> Have you got any evidence of that? How can we collect evidence?

Currently every project behind bundlers use either `import` and let the bundler solve the kind or just `require` which works out of the box.

Everything a bundler or synchronous loader, needs to do with `import.meta.require` is to drop `import.meta.` and keep `require`, if it's targeting CJS and transforms all pseudo ESM into CJS, or do nothing if the env is ESM and the native `import.meta.require` is used.

This is a migration path that requires zero effort for users and tiny effort for bundlers.

Anything else that needs user-land solution won't be as simple, straight forward, and effective as `import.meta.require` and bundlers won't be able anymore to understand where and how files should be loaded because every logic to require CJS will be potentially different and harder to analyze/understand.

Goodbye tree shaking, code splitting, and ease of parsing if we push out a user-land based, potentially always different per project/company, module loader.

This is my _logical_ evidence.

zenparsing:
@bmeck Thanks for the counter-points, we'll definitely need to weight them.

I'm not sure why we're discussing extending Node core here. Shouldn't that be out of scope?

The only other comment I'd like to make is that `import.meta` is designed specifically for this kind of module-context-specific stuff. Having users import a function from core and supply the context via a "url" is a bit like circling the house to go in the front door.

ljharb:
@mcollina it’s worth noting that https://npmjs.com/resolve (8.6m/week) achieves that functionality without using undocumented APIs.

bmeck:
@WebReflection 

> Currently every project behind bundlers use either import and let the bundler solve the kind or just require which works out of the box.

I'm not sure this really relates as bundlers implement their own form of `require` instead of using the Node one, and they also convert `import` to their own bundling format. The discussion here is mostly tailored around how to let users write as much code in ESM that is unchanged when passed to both node and potentially other environments.

> Everything a bundler or synchronous loader, needs to do with import.meta.require is to drop import.meta. and keep require, if it's targeting CJS and transforms all pseudo ESM into CJS, or do nothing if the env is ESM and the native import.meta.require is used.

This tailors the experience to the bundler, I am more concerned with allowing the user to be encouraged down a path of maximal compatibility than pleasing tool makers.

> This is a migration path that requires zero effort for users and tiny effort for bundlers.

Bundlers could easy boilerplate out the bundle example I have above. This doesn't seem to discount that nor does it address that users must also migrate all code to `import.meta.require` instead of `require`, so there is migration effort on both. I have little concern for overall implementation effort on bundlers for `createRequireFunction` as there are plenty of ways to do bailouts and tree shaking without relying on `import`/`export` and using variable tracking of different kinds.

> Anything else that needs user-land solution won't be as simple, straight forward, and effective as import.meta.require and bundlers won't be able anymore to understand where and how files should be loaded because every logic to require CJS will be potentially different and harder to analyze/understand.

I don't understand how this relates when bundlers could just alter output formats. This also seems to be a claim about CJS being hard to analyze that is applied to both `import.meta.require` and `createRequireFunction`. If this could be expanded upon and explained in a more concrete way, that would potentially show why this point is relevant.

> Goodbye tree shaking, code splitting, and ease of parsing if we push out a user-land based, potentially always different per project/company, module loader.

This doesn't make sense.

> This is my logical evidence.

I'd once again encourage you to not claim your own point as a conclusion at the end of your statements. It often creates both friction and confusion. In this case, mostly confusion on my part.

WebReflection:
> I'd once again encourage you to not claim your own point as a conclusion at the end of your statements.

At least I always state it makes sense and it's logical for me, I don't claim my personal point of view as the only truth out there, which is how I often interpret your statements.

Your post is subjective as much as mine, if you want, and specially the static analysis one is a deviation of how `import.meta.require` would shine instead of a in-house, per user, loader.

So, I guess we are both confused on each other statements, but if you want a concrete example on how I can statically find `import.meta.require` even via `grep` and never be able to understand anything via a `createRequireLoader` I'd be happy to write down that code.

WebReflection:
P.S. it would be very valuable if bundlers related developers would say something about this too, i.e. @TheLarkInn

michael-ciniawsky:
Bundlers mainly care about the `CallExpression` or keyword when parsing the code and apply there own handling and logic afterwards anyways. `import.meta.require` is likely simpler to parse here (on the edges, whatever the are) since `createRequire` needs to be tracked from the call to the import to know where it originally came from

zenparsing:
@bmeck I just mean that we should try to resist the temptation to justify things one way or the other by supposing that we're solving problems that are out of scope.

As far as the general issue goes, my opinion is that the benefits of "createRequire" are pretty well non-existent considering that the "require" problem (CJS-in-the-browser) was solved a long time ago by bundlers.

Let's say that I want to create an ESM entry point for a CJS package with a default and named exports. 

With `createRequire` it might look like this (assuming that `createRequire` can handle a URL):

```js
// An ESM entry point for my CJS package
import { createRequire } from 'module';
const exports = createRequire(import.meta.url)('./index.js');
export default exports;
export const { a, b, c } = exports;
```

With `import.meta.require` it looks like this:

```js
// An ESM entry point for my CJS package
const exports = import.meta.require('./index.js');
export default exports;
export const { a, b, c } = exports;
```

Either way, it's not that big of a deal. As a user, I would think the first is OK, but I would like the second better. Bundlers are going to have an easier time with `import.meta.require`.

For me, the scale seems to tip slightly toward `import.meta.require`. It's just really intuitive and "feels" right.

WebReflection:
Thanks @michael-ciniawsky , that underlines my statement: `import, meta.require` is zero effort for users (no in-house loaders, no best practice or error handling done right to teach) and little effort for bundlers (at least littler).

guybedford:
Imagine teaching someone new to Node.js how to use modules in Node. They're writing an ES module, and want to run it in both the browser and Node.js, and they are loading a third-party package, which they've installed from npm, but that also available on a CDN for the browser.

How would you explain `import.meta.require` to them? How would you explain to them how they can run this ES module in the browser? Does that feel intuitive to this user? Or is it just because we've been writing `require` for so many years?

devsnek:
@zenparsing build tooling shouldn't be replacing calls to makeRequireFunction, it should be replacing `import 'module'`. i think your entire argument is invalidated by this.

zenparsing:
> How would you explain import.meta.require to them?

In this hypothetical example, why would they be using `import.meta.require`? If they are writing an ESM package then why would they need it?

@devsnek 

Whatever way a build tool chooses to deal with CJS-isms, it's not going to make a difference to the user. To be clear, my argument is basically that we shouldn't be introducing friction based on:

- Discouraging users from using it (they'll use it if they need it, not because they are "tempted")
- That "createRequire" has benefits in the browser (bundlers deal with that)

WebReflection:
@guybedford assuming you are asking me ...

> Imagine teaching someone new to Node.js how to use modules in Node

I'd teach to simply use CommonJS because ESM state is a mess right now. That's me teaching, but answering further ...

> They're writing an ES module, and want to run it in both the browser and Node.js

They should just use ESM then, as it works on the only environment that supports ESM already: the browser

> and they are loading a third-party package, which they've installed from npm, but that also available on a CDN for the browser

If this package is published as CJS I'd teach them how to use bundlers, because on the Web nothing else would work.

If this package is published in ESM and we live already in 2080 where ESM landed in Node (sorry for the stretch) I'd have nothing to teach, hoping by that time ESM in node accepts remote URLS too like the browsers do.

> How would you explain import.meta.require to them?

If I have to explain `import.meta.require` to them it means they need to import a, by that time, legacy CJS module, instead of an ESM one, so I'd tell them that like Python used `from __future__ import module`, which is not a long term solution to modules, in NodeJS you can `from __past__ import module` which is CJS.

That means, I'd explain that they always have to `import stuff from 'module'` but if the module is not compatible with legacy they need to use `const stuff = import.meta.require('module')`.

I'd also explain that bundlers would solve this automatically for them, but if they want to publish an executable or cli, or any production code that shouldn't trust bundlers, then they can easily use _is-node-legacy_ CLI tool to understand if a module should be loaded as ESM or as CJS.

I see zero friction/issues for users, and little effort for bundlers 🤷‍♂️ 

> Does that feel intuitive to this user?

It's one thing to explain, instead of explaining that they should use ESM to import something that is used to require CJS but that needs to be done in a way that makes sense or shenanigans happens.

So I am pro UX here, not sure about you.

> is it just because we've been writing require for so many years?

I've used XMLHttpRequest for 10 years and fetch came after ... I am still fully capable of understanding where fetch is superior and where XHR is better (i.e. sync requires or easy abortion ability).

michael-ciniawsky:
> Thanks @michael-ciniawsky , that underlines my statement: import.meta.require is zero effort for users (no in-house loaders, no best practice or error handling done right to teach) and little effort for bundlers (at least littler).

Yes and no, since bundlers parse the JS it is definitely doable to use `createRequire` but it likely envolves more work to handle all cases of potential `createRequire` usage. But do bundlers need to care about where `createRequire` is coming from in most cases or at all? It's really mainly about finding `require/import/import()/new Worker() (and variants)` AST nodes and then it's bundler logic taking place from there. There are variants of course if `createRequire` is used e.g by 'directly' calling it on the right side of assignment (see one of @zenparsings examples above) but that mainly just adds `createRequire` as a potential `CallExpression` name to the list (`CallExpession { name: 'createRequire', ...}`)

```js
const require = import.meta.require
const require = createRequire(...)

const dependency = require('path/to/dependency.js') // <= CallExpression { name: 'require', ...}
```

guybedford:
@zenparsing 

> In this hypothetical example, why would they be using import.meta.require? If they are writing an ESM package then why would they need it?

The issue here is all of the utility code for browsers that exists as CommonJS on npm. Say I want to use a MIME parsing package etc.

The result is that if I use one of 700,000 packages written in CommonJS, my module will no longer work in a browser.

This puts a pressure on:

1. Every one of those 700,000 packages to upgrade to ES modules (think of all the PRs / issues that willl be posted, and while yes they would be anyway, this would magnify this pressure by orders of magnitude), and many of these are authors who would consider their work complete, and are not looking to upgrade.

2. Users now have a strong incentive to select dependency packages that support ES modules. This means things like ES module badges on packages when browsing the registry, and the inevitable disappointment when a useful package doesn't support ES modules and requires you, as a user, to change your entire workflow to ensure it can support `import.meta.require` (eg we use TypeScript to compile to AMD and now need to add a custom Babel transform into our build to convert import.meta.require in to AMD-style CommonJS require). Or whatever myriad of build approaches are out there. The act of requiring a CommonJS package now mandates a new workflow and build system, while the act of requiring a CommonJS package with transparent interop is exactly what users are doing already, and have already set up their builds to deal with.

We need to accept that there are costs on workflows in various directions here, and make sure to weigh them up with our eyes open.

WebReflection:
> The result is that if I use one of 700,000 packages written in CommonJS, my module will no longer work in a browser.

Why would that be ? Right now everything works, why wouldn't it work all of a sudden ?

> Users now have a strong incentive to select dependency packages that support ES modules

Meaning they are all behind bundlers 'cause they are lazy enough to let th bundler distinguish.

In such scenario, for those few that needs to consciously require a CJS module in ESM, let it be simple.

Every other scenario you see as a problem is nothing different from today scenario where bundlers, and developers publishing dual modules like me, already faced, and solved.

All we need now is a way to explicit import from the past, which is CJS.

zenparsing:
@guybedford Are you suggesting that we ought not have `import.meta.require` or `createRequire`, and instead should be able to import from CJS directly?

That conversation might be a bit off-topic here, but I'll just breifly state that it's a non-starter for me due to the named-imports problem.

To your point about existing packages having to be patched to support (node-native) ESM, that doesn't seem fatal to me, as long as it's easy enough to do. Let's not underestimate the ability of the ecosystem to adapt, particularly if we give authors enough lead time.

devsnek:
i'm not comfortable shipping an implementation that needs bundlers to be usable.

Janpot:
Can't `import.meta.require` just be shimmed on top of `createRequire` with a loader? For people that prefer the convenience?

devsnek:
@Janpot 👍 We could expose `import.meta` creation hooks.

WebReflection:
@devsnek bundlers are the present, and the short term future, there's no CommonJS interoperability whatsoever without those, 'cause CommonJS is not a standard, and it never shipped on the Web.

If you create projects that need CommonJS modules, you either use bundlers, or you cannot use those modules, on the Web, while if you do so for NodeJS, there's nothing to worry about, 'cause you won't need bundlers to use `import.meta.require`.

TL;DR your concern makes no sense on the Web 'cause it's mandatory to use bundlers there, for CJS.

guybedford:
> @guybedford Are you suggesting that we ought not have import.meta.require or createRequire, and instead should be able to import from CJS directly?

I don't mind a createRequire, as there are lots of things you can do in Node that don't work in browsers, and I'm not entirely against import.meta.require, I'm just saying that if we put the stamp of sugar on `import.meta.require`, we had better be sure we understand and appreciate the implications throughout the ecosystem, these are the only points I'm just trying to bring up on this.

> That conversation might be a bit off-topic here, but I'll just breifly state that it's a non-starter for me due to the named-imports problem.

This is why I created the dynamic modules proposal- https://github.com/guybedford/proposal-dynamic-modules. Would be interested to hear your feedback here.

> To your point about existing packages having to be patched to support (node-native) ESM, that doesn't seem fatal to me, as long as it's easy enough to do. Let's not underestimate the ability of the ecosystem to adapt, particularly if we give authors enough lead time.

Well the thing is it just takes one useful package not to upgrade to degrate an entire workflow from being "modern". I don't know what that will look like exactly though, but we must try to think this through. I just imagine a lot of swearing when I think about it :P

bmeck:
@Janpot you could use a loader, or put the following at the top of any given file manually:

```mjs
import {createRequireFunction} from 'module';
import.meta.require = createRequireFunction(import.meta.url);
```

However, that would still mean migration of code from `require` to `import.meta.require`.

And people would probably want to DRY that up and/or prevent code modification that isn't needed that up so you end up with:

```mjs
var require = import.meta.require
```

Being likely in both scenarios, so bundlers will still have to track that `require` for that scenario in both scenarios.

Luckily for both `import.meta.require` and `createRequireFunction` bundlers already do variable tracking and don't just use call sites as you can see with the error on https://codesandbox.io/s/6y5vzp5xr

Janpot:
Yep, I understand. I just wanted to add another point in favour of `createRequire`. As an end-user of node, I prefer `require` in ESM being done in a way that doesn't pollute the ESM primitives if it is perfectly shimmable with a loader.

bmeck:
@Janpot even if we don't consider this "pollution", I still have concerns since you can *always* put:

```mjs
// this could be safely at the top of file shipped to browsers if the file doesn't use `require()`
// even though it doesn't do what you might expect
var require = import.meta.require;
```

At the top of files even if `require` becomes undefined instead of the expected function. Meaning, it could be added as a boilerplate even when `import.meta.require` doesn't exist since `import.meta.require` doesn't fail early. Therefore we could see this line appear in ecosystems even without the function existing just as precaution or habit. Somewhat similar to the times I see `"use strict"` at the top of ESM source text.

Janpot:
> Somewhat similar to the times I see "use strict" at the top of ESM source text.

But `"use strict"` in ESM would strictly be a no-op.
Couldn't `var require = import.meta.require;` potentially overwrite `require` in global scope?

ljharb:
Modules never run in the global scope, nor anything a bundler bundles.

However, it’s syntax, so it would fail to parse in non modern browsers, unlike a pragma.

zenparsing:
I think this conversation has been helpful. Personally, my takeaways are:

- The bundler argument isn't overly strong, since bundlers will probably want to do some variable tracking either way. It will probably be a little bit easier to track usage of `import.meta.require`.
- The polyfill-ability of "createRequire" is probably not much of an advantage, since the ecosystem has moved toward static analysis and bundling.
- I'm still not sure what to make of the "tech-debt" argument, since Node will always need to support CJS in one way or another; both "createRequire" and `import.meta.require` are equally "legacy".

Provided that you can do `createRequire(import.meta.url)` (can it accept either a path or url string?), it seems tolerable. I imagine that from a user's point of view it would also seem like node is being intentionally difficult.

bmeck:
> Provided that you can do createRequire(import.meta.url) (can it accept either a path or url string?), it seems tolerable. I imagine that from a user's point of view it would also seem like node is being intentionally difficult.

https://github.com/nodejs/node/pull/19360/files#diff-76195ce57689942222a27f0dbda6d3b7R749 has it accepting both.

zenparsing:
Accoding to that PR it looks like you have to do:

```js
import { createRequire } from 'module';
const require  = crequireRequire(new URL(import.meta.url));
```

I think that creating a URL instance from import.meta.url pushes it over the edge into incantation.

bmeck:
@zenparsing ah, well filepaths and URL strings are not mutually exclusive and have collisions. If it only took URL strings would that be sufficient?

CC: @devsnek

michael-ciniawsky:
I personally think the most sensible approach for now would be to add a `createRequire` method to `module` first, since this is potentially useful in CJS aswell and `import.meta.require` is mainly 'sugar' for `createRequire` in ESM which can always be added/switched to later on e.g after a Developer Survey/RFC

devsnek:
@bmeck @zenparsing I have use cases for createRequireFunction outside of `import.meta.url` but feel free to comment on the PR and we can go more in depth on this.

MylesBorins:
+1 to moving forward with create require and revisiting this as part of a
larger proposal

zenparsing:
Just throwing a possibility out there:

A design principle we could use here is that we should provide ESM module authors with pretty much the same old Node API that they currently have. So:

- All globals are the same
- All module-specific free variables are copied over to "import.meta"

So we would have `import.meta.dirname`, `import.meta.filename`, and `import.meta.require`. I don't think "module" would make any sense at all, though.

Perhaps controversial, but at least it would be consistent!

bmeck:
@zenparsing maybe later, but I feel like `import.meta.url` serves to encourage people to migrate to using that which will work better across environments and won't introduce more API surfaces. They would still need to port code to use `import.meta.*` for anything. The familiarity of the API names would match, but it wouldn't make things necessarily better since it would still require breaking changes to code. In addition, if we ever support non-file loading (such as directly from memory) the `filename` and `dirname` API would get confusing (should it throw on access?).

SMotaal:
I'd like add my own take on things...

```js
import { createRequire } from 'module';
import.meta.require = createRequire(import.meta.url);
```

In fact, `createRequire` can have some sibling concepts:

```js
import { fromMeta } from 'module'; // ie Module.fromMeta(…)
const { __dirname, __filename, require } = fromMeta(import.meta);
```

- This obviously is no browser module as is, unless replace `from 'module'` with from `from '/that-awesome-platform/module.mjs'` or something, easy RegExp fix in a service worker.
- Bundlers face one new challenge.
- Users see one new API and rely on the 'module' they always loved in Node.js.

I really believe confusion can be avoided best in the fewest lines of code only if they are mostly familiar concepts, with the one or as few new API's as possible.

demurgos:
@SMotaal 
I really like your proposition to create all the CJS metavariables explicitly from the URL.
Even if `require` is at the core of this issue, dealing with `__filename` and `__dirname` is also very important. Converting URLs to paths is tricky. Even among the members of the module working group, some have posted snippets with invalid conversions.

A minor note: one of the features listed for Node's ESM is support for modules without a file URL. This function should be aware of this case.

zenparsing:
Seems reasonable.

Is the guiding principle here that we should not put anything on `import.meta` that is non-portable?

@domenic What are your thoughts from a web point of view? Should Node resist the temptation to put anything "node-esque" on `import.meta`, such as "filename" and "dirname", or is it all fair game for host environments?

devsnek:
@zenparsing 

we can put node-esque things on `import.meta`. the problem with `dirname` and `filename` is twofold.

1) not all files will have anything valid to put for `dirname` and `filename`
2) you still have to then refactor all your `__dirname` and `__filename` nonsense.

IMO these values are so underused that there isn't really a need to put them so accessible. People who need these values are usually doing a lot of weird meta stuff anyway so an additional call to `path.dirname` or `getFilePathFromURL` is fine.

domenic:
Putting host-environment specific stuff on import.meta is fine, and even encouraged. We still hope to put something like import.meta.scriptElement on there in browsers. We can of course coordinate across environments, but environments have different requirements that will definitely lead to non-overlap.

I strongly suggest not using context-dependent modules to import metadata and other such functions. I wrote about this before at https://gist.github.com/ceejbot/b49f8789b2ab6b09548ccb72813a1054#gistcomment-2321028

guybedford:
@domenic `createRequire` as discussed here is not a context-dependent import, it is a factory function that would be passed `import.meta.url`, within the bounds of the module system semantics.

WebReflection:
just pointing out that others favorited `import.meta.require` before and reasons look always the same:
https://gist.github.com/ceejbot/b49f8789b2ab6b09548ccb72813a1054#gistcomment-2330869

guybedford:
@WebReflection we are all well aware that `import.meta.require` is a nice pattern for Node integration in isolation, the concerns are browser and tooling compatibility (as well as the inability for node bins to work cross platform).

zenparsing:
@domenic Thanks (and apologies if you now get pinged on this thread!)

@devsnek 

We'll have the same problem with any of these variations, I think. If `getFilePathFromURL` is given a non-file URL it would have to throw or something, and I imagine that as @bmeck suggested, `import.meta.dirname` would have to throw if it's not a file-system module.

I suppose the principle at play here is that properties on `import.meta` are less obviously non-portable. The counter-argument is that `import.meta` is **by definition** non-portable. From that point of view, the fact that we have overlap in terms of "url" is somewhat misleading.

devsnek:
@zenparsing loader assumes you're just on the filesystem because this behaviour anywhere else is objectively terrible, which is why i don't want it by default in node.

zenparsing:
> A future where everyone wraps accessing properties from import.meta in try/catch because we couldn't design a proper system of integration sounds terrible to me.

Don't users have to do the same thing with `getPathFromURL`? I suppose that by requiring the indirection from "url" we would discourage (via friction and annoyance) non-portable code. I guess I don't follow how that helps anything though. If I need to get a dirname, I'm going to get it, whether Node makes me write one line or two or five. Either way, I'm writing non-portable code.

There's another tool for discouraging certain patterns: linters. Linters can warn if you're using `import.meta.dirname` in a portable codebase. Developers have to use that strategy for globals and things already.

michael-ciniawsky:
`import.meta.node` ?

```js
const { require, __dirname, __filename }  = import.meta.node
```

```js
// or (depending on individual feature usage and use cases)
const require = import.meta.require
const { __filename, __dirname } = import.meta.node
```

ljharb:
Linters can also warn against usage of any object properties, any global variables, and any imports or requires - `import.meta.url` just makes the job of detection slightly easier for the < 100 (< 1000?) people that will ever be writing bundler and linter code directly.

devsnek:
@michael-ciniawsky we already own the `import.meta` object, we can put whatever we want on it. again this isn't a problem of putting environment specific properties on `import.meta`.

guybedford:
It is possible to write an implementation of `url` that works in the browser, so that code that does `import { getPathFromFileURL } from 'url'; import fs from 'fs'; fs.readFile(getPathFromFileURL(import.meta.url))` could be portable using a form of virtual fs.

devsnek:
@guybedford `browserify` (and by extension webpack) both use npmjs.org/url to polyfill in browser. as long as that module is updated, i don't see why that wouldn't work.

michael-ciniawsky:
What is the concrete problem then ?

Why is something like `getPathFromFileURL` needed ?

```js
import { getPathFromFileURL } from 'url'; 
import fs from 'fs'; 

fs.readFile(getPathFromFileURL(import.meta.url))
```

Why not

```js
const { pathname } = new URL(import.meta.url)

fs.readFile(pathname + 'file.ext', cb)
```

What is the difference ?

devsnek:
@michael-ciniawsky https://github.com/nodejs/node/pull/22506/files#diff-dd182d238def46a3af0d112f617e8838R883

michael-ciniawsky:
🙄 

```
const url = import.meta.url

const file = path.from(url)
```

@devsnek thx

jasnell:
The difference is that `fs.readFile(pathname + 'file.ext', cb)` would not produce a valid path across platforms. File URL pathname values are very different from filesystem paths and must be translated in a platform specific way in order to be used.

michael-ciniawsky:
Can't the various `node` API's which currently accept a `path` either be restricted from URL usage (where appropiated) or nomalize (File) URL's themselves instead of having users to deal with that in (clunky ways) within userland ? 

### 🤢 

```js
import url from 'url'
import path from 'path'

const p = path.from(import.meta.url)
const p = url.getPathFromFileURL(import.meta.url)
// etc etc...

fs.readFile(p, cb)
```

### 😃 

```js
const file = url || path + 'file.ext'

fs.readFile(file, cb)
```

But I think this should ideally be a separate topic and issue, I'm for e.g currently confused why && where File URL usage is even needed since e.g the `fs` example in particular would be `node` only anyways and a `const { __dirname, __filename } = import.meta.x` should be sufficient 🤷‍♂️

jasnell:
They already do, to an extent. You can pass a File URL to `fs.readFile()` and it will convert it for you. What you have to be careful about, however, is passing just the pathname segment or using concatenation to modify the input. You have to pass in File URL object itself in order for it to work.

jasnell:
So if `import.meta.url` returns a URL object, you could just do `fs.readFile(import.meta.url)` .... If it returns a string you would do `fs.readFile(new URL(import.meta.url))`

devsnek:
and for the record, `import.meta.url` must be a string. URL objects are mutable.

https://url.spec.whatwg.org/#url-apis-elsewhere

SMotaal:
@bmeck I think from an interop standpoint, filename and dirname are not file-system specific, they could be URL derived outside of node with roughly `fileURL.replace(/^.*[/]([^/?#]*?)(?:[?#].*|)$/, '$1')` and `new URL('.', fileURL).pathname` not withstanding which of RegExp or URL would be the better here

But ultimately, the fallback is `undefined`, but then the question is do we expect people to want to do something like `'filename' in meta` or just `const {filename = 'fallback'} = meta` (I use meta to avoid suggesting endorsement at this point).

MylesBorins:
Explicitly adding blocked to make it clear there is not an expectation for this to land right now, but rather this is a archive and documentation of the work on import.meta.require so it doesn't need to be done again in the future.

Open to closing the PR if people think that is a better appraoch

SMotaal:
@MylesBorins Can you clarify why "blocked" makes its was to this PR? Is that the direct result of [module: add createRequireFunction method](/nodejs/node/pull/19360)?

Can you help me connect the dots please.

MylesBorins:
@SMotaal as we have agreed to focus on the minimal kernel there is no reason to land this right now. I don't think it makes sense to close it and lose the work. As such the blocked label clearly communicates there is not a current intent to merge this

SMotaal:
Awesome… makes sense.

MylesBorins:
This is not likely landing anytime soon, we can revisit another time

PR #12, opened 2018-10-24
esm resolver spec and implementation refinements

guybedford:
Moved from original posting at https://github.com/nodejs/modules/pull/207:

In line with the current plans for Phase 2 I've put together a draft resolver algorithm here that includes the new properties from the minimal kernel implementation.

Having an easy-to-read algorithm will allow tools and bundlers to follow the resolver implementation, just like with the CommonJS resolver.

The goal is, like the minimal kernel, to start with what we have agreed on, and to expand this over time as we can reach consensus.

This one brings up the discussions around the package.json main handling.

Would be nice to discuss this at the meeting today if there is time, although I understand it is late notice to read up on the details here.

Readable link: https://github.com/nodejs/ecmascript-modules/blob/esm-resolver-spec/doc/api/esm.md#resolution-algorithm


doc/api/esm.md
```diff
@@ -149,3 +149,97 @@ fs.readFileSync === readFileSync;
...
+1. _ESM_RESOLVE(@pkg/name, file:///path/to/project)_ -> _file:///path/node_modules/@pkg/name/index.mjs_ (otherwise, if it exists)
+1. _ESM_RESOLVE(@pkg/name, file:///path/to/project)_ -> _file:///node_modules/@pkg/name/index.mjs_ (otherwise, if it exists)
+1. _ESM_RESOLVE(@pkg/name, file:///path/to/project)_ -> _Not Found_ (otherwise)
+
+### Resolver Algorithm
+
+The algorithm to resolve an ES module specifier is provided through _ESM_RESOLVE_:
+
+ESM_RESOLVE(specifier, parentURL)
+> 1. If _specifier_ starts with _"/", _"./"_ or _"../"_ then,
```
guybedford: Comment from @jkrems - 

Should we copy the beginning from the HTML spec even though the observable behavior should be the same?

Apply the URL parser to specifier. If the result is not failure, return the result.
If specifier does not start with the character U+002F SOLIDUS (/), the two-character sequence U+002E FULL STOP, U+002F SOLIDUS (./), or the three-character sequence U+002E FULL STOP, U+002E FULL STOP, U+002F SOLIDUS (../), return <our custom logic>.
Return the result of applying the URL parser to specifier with base URL as the base URL.
That ways it's a bit easier to tell where we deviate from what would happen in a browser.

guybedford: @jkrems certainly we could do this, although personally I like that the NodeJS CommonJS resolver is very readable. Hopefully we can strike a better mark on correctness while retaining readability here.

guybedford:
Readable link: https://github.com/nodejs/ecmascript-modules/blob/esm-resolver-spec/doc/api/esm.md

doc/api/esm.md
jkrems: To clarify: My main point was about the order of operation, not necessarily the precise wording. E.g. "if valid URL, return" being the first step.

guybedford: Because there are no valid URLs beginning with `./`, `../` or `/` I considered this an irrelevant difference (and also it gives better performance the other way around), but we can certainly switch this to follow WhatWG.

jkrems: Yeah, I thought the same thing but I had to spend a minute or two thinking if there are cases where it would make a difference. So I'd prefer calling out our intend ("does the same thing as the browser for these") and handle optimizations ourselves if we determine that turning things around is faster.

doc/api/esm.md
```diff
@@ -149,3 +149,99 @@ fs.readFileSync === readFileSync;
...
+
+### Features
+
+The resolver has the following properties:
+
+* FileURL-based resolution as is used by ES modules
+* Relative and absolute URL resolution
+* No default extensions
+* No directory lookups
+* Bare specifier package resolution lookup through node_modules
```
jkrems: I'm not sure if we should take on the historical burden of `node_modules` in the default implementation. It's inefficient, hard to predict (e.g. it might load random things from a user's home directory), and enshrines a directory layout that many people (myself included) are trying to get away from.

Should we maybe split this into scoped resolution and an example for how the underlying data for scoped resolution might be calculated from a directory tree?

ljharb: I'm confused, how is it hard to predict?

That directory layout isn't something that can be trivially nor any time soon gotten away from.

doc/api/esm.md
guybedford: Sure, I've switched around the order to match.

guybedford:
@devsnek thanks, always open to spec style suggestions, I've added those.

doc/api/esm.md
```diff
@@ -149,3 +149,101 @@ fs.readFileSync === readFileSync;
...
+> 1. Return the result of **PACKAGE_RESOLVE**(_packageName_, _packagePath_, _parentURL_), or propagating any errors thrown.
+
+**PACKAGE_RESOLVE**(_packageName_, _packagePath_, _parentURL_)
+> 1. Assert: _packagePath_ is either empty, or a path without a leading separator.
+> 1. Assert: _packageName_ is a valid package name or scoped package name.
+> 1. Note: Further package name encoding validations can be added here.
+> 1. If _packagePath_ is empty and _packageName_ is a Node.js builtin module then,
+>    1. Return _"node:${packageName}"_.
+> 1. Set _parentURL_ to the parent folder URL of _parentURL_.
+> 1. While _parentURL_ contains a non-empty _pathname_,
```
sendilkumarn: Is this _packagename_ instead of _pathname_ ?

doc/api/esm.md
```diff
@@ -149,3 +149,101 @@ fs.readFileSync === readFileSync;
...
+>    1. Let _packageURL_ be equal to _"${parentPath}/node_modules/${packageName}"_.
```
sendilkumarn: ${_ParentUrl_}/....

doc/api/esm.md
```diff
@@ -149,3 +149,101 @@ fs.readFileSync === readFileSync;
...
 [`module.createRequireFromPath()`]: modules.html#modules_module_createrequirefrompath_filename
 [ESM Minimal Kernel]: https://github.com/nodejs/modules/blob/master/doc/plan-for-new-modules-implementation.md
+
+## Resolution Algorithm
+
+### Features
+
+The resolver has the following properties:
+
+* FileURL-based resolution as is used by ES modules
```
bmeck: this doesn't cover the builtin libraries? maybe we should call out that this algorithm is for non-builtins somewhere?

doc/api/esm.md
```diff
@@ -149,3 +149,101 @@ fs.readFileSync === readFileSync;
...
+* Relative and absolute URL resolution
+* No default extensions
+* No directory lookups
+* Bare specifier package resolution lookup through node_modules
+
+### Package main
```
bmeck: wouldn't this section mean not supporting `package.json` in arbitrary directories? If so I'd rather hold of until we have discussions on this change.

doc/api/esm.md
```diff
@@ -149,3 +149,101 @@ fs.readFileSync === readFileSync;
...
+Package resolution:
+
+1. _ESM_RESOLVE("pkg/x", "file:///path/to/project")_ -> _file:///path/to/node_modules/pkg/x_ (if it exists)
+1. _ESM_RESOLVE("pkg/x", "file:///path/to/project")_ -> _file:///path/node_modules/pkg/x_ (otherwise, if it exists)
+1. _ESM_RESOLVE("pkg/x", "file:///path/to/project")_ -> _file:///node_modules/pkg/x_ (otherwise, if it exists)
+1. _ESM_RESOLVE("pkg/x", "file:///path/to/project")_ -> _Not Found_ (otherwise)
+
+Main resolution:
+
+1. _ESM_RESOLVE("pkg", "file:///path/to/project")_ -> _file:///path/to/node_modules/pkg/index.mjs_ (if it exists)
```
bmeck: shouldn't this first search within `/path/to/project/node_modules`????

doc/api/esm.md
guybedford: Builtins are actually supported, I will add a note.

doc/api/esm.md
guybedford: This is intended to be pathname.

doc/api/esm.md
guybedford: Yes, thanks.

doc/api/esm.md
guybedford: Ok, I've added non-package mains as well for now. We can discuss further though.

doc/api/esm.md
guybedford: Yes, thanks.

guybedford:
@bmeck thanks for taking a look - all feedback has been updated in the spec.

guybedford:
@nodejs/modules Initially I attempted to remove directory indexes from this PR entirely, as in the current minimal implementation, but this again results in the friction that `import 'pkg'` will apply the main but `import '/path/to/node_modules/pkg'` won't apply the main which just seems like it will break a lot of things having this inconsistency.

The same applies for `pkg/x` where `x` is a directory with a main. Where we then have the issue that if `/path/to/node_modules/pkg/x` will work, then we should have `pkg/x` also work here for consistency.

Note that no extra statting is incurred by this, and it gives us maximum backwards compatibility.

But what it comes down to is that folder mains almost impossible to deprecate, unless we want to break lots of things.

doc/api/esm.md
```diff
@@ -149,3 +149,109 @@ fs.readFileSync === readFileSync;
...
+Main resolution:
+
+1. _ESM_RESOLVE("pkg", "file:///path/to/project")_ -> _file:///path/to/project/node_modules/pkg/index.mjs_ (if it exists)
+1. _ESM_RESOLVE("pkg", "file:///path/to/project")_ -> _Not Found_ otherwise, even if _file:///path/to/node_modules/pkg/index.mjs_ exists.
+1. _ESM_RESOLVE("file:///path/to/project")_ -> _file:///path/to/project/index.mjs_ (if it exists)
+1. _ESM_RESOLVE("file:///path/to/project")_ -> _Not Found_, if _file:///path/to/project/index.mjs_ does not exist.
+
+### Resolver Algorithm
+
+The algorithm to resolve an ES module specifier is provided through _ESM_RESOLVE_:
```
xtuc: nit: `_ESM_RESOLVE_` -> `_ESM_RESOLVE` or the other way arround.

guybedford: Sure, I've moved to just italicizing the arguments.

guybedford:
I've gone ahead and updated this PR with an implementation of the spec as written as well.

In the process the error handling has also been improved to include the parent importer path when throwing not found errors, as well as providing the most relevant version of the specifier in the resolution process.

For example if resolving `import 'x/y'` where the package 'x' exists at `node_modules/x` but without a `y`, the error thrown will read:

```
Cannot find module '/path/to/node_modules/x/index' imported from /path/to/x.mjs
```

This is also a consequence of no longer implementing the node_modules fallback for missing subpaths.

In addition error message consistency is provided to ensure all NOT FOUND errors have the same initial description structure of `Cannot find module '%s' imported from %s`.

Apart from that the rest just implements the spec directly.

src/module_wrap.cc
```diff
@@ -467,104 +467,108 @@ std::string ReadFile(uv_file file) {
...
+      DescriptorType check = CheckDescriptor(url.ToFilePath());
+      if (check == FILE) {
+        return Just(url);
+      } else if (check == DIRECTORY) {
+        Maybe<URL> main_url = PackageMainResolve(url);
+        if (!main_url.IsNothing()) return main_url;
       }
-      return check;
     } else {
-      // TODO(bmeck) PREVENT FALLTHROUGH
```
guybedford: Note this TODO is fixed :)

guybedford:
In the name of a minimal implementation I've made the following further changes here:

1. Remove folder mains - these can always be added back
2. Removed the `index.mjs` main to simply have no main currently - rather splitting this out as further work.
3. Unified error handling to have all resolver errors be ERR_MODULE_NOT_FOUND instead of having ERR_MODULE_LEGACY, ERR_MISSING_MODULE, MODULE_NOT_FOUND all meaning the same thing.

jkrems:
+1 to merging this to allow further iteration.

guybedford:
Further to the discussion in today's meeting I've removed all main handling and parsing from this PR.

@nodejs/modules I would be very grateful if we can merge this before the next meeting so that we can consider PRs against this then. If everyone could take a moment to review I'd be very grateful to get this spec in!

guybedford:
@giltayar hopefully this allays your concerns here?

doc/api/esm.md
```diff
@@ -146,6 +142,115 @@ fs.readFileSync = () => Buffer.from('Hello, ESM');
...
+* FileURL-based resolution as is used by ES modules
+* Support for builtin module loading
+* Relative and absolute URL resolution
+* No default extensions
+* No folder mains
+* Bare specifier package resolution lookup through node_modules
+* Avoids the package fallback which breaks package isolation. This is the
+  problem that an import to `x/y.js` will keep checking subsequent node_modules
+  even if a package "x" is found without a "y.js". This is done through package
+  name format detection.
```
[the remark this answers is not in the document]
guybedford: @giltayar I think you are still looking at the previous PR version here, but glad to discuss this further.

Perhaps it would help if you clarify your concerns around the package name parsing?

It is fairly well-established at this point in the game and I personally find it a much more well defined model than supporting arbitrary modules in node_modules (eg `require('x.js')` will load node_modules/x.js.

The package fallback problem is not a large issue, but it's a nice to fix.

doc/api/esm.md
```diff
@@ -146,6 +142,115 @@ fs.readFileSync = () => Buffer.from('Hello, ESM');
...
+>       1. Throw a _Invalid Package Name_ error.
+>    1. Set _packageName_ to the substring of _packageSpecifier_
+>       until the second _"/"_ separator or the end of the string.
+> 1. Let _packagePath_ be the substring of _packageSpecifier_ from the
+>    position at the length of _packageName_ plus one, if any.
+> 1. Assert: _packageName_ is a valid package name or scoped package name.
+> 1. Assert: _packagePath_ is either empty, or a path without a leading
+>    separator.
+> 1. Note: Further package name validations can be added here.
+> 1. If _packagePath_ is empty and _packageName_ is a Node.js builtin
```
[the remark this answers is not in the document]
guybedford: CommonJS does actually support `node_modules/fs/` being looked up via `require('fs/...')`.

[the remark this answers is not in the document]
guybedford: Package name parsing would help with this potentially too.

doc/api/esm.md
```diff
@@ -146,6 +142,115 @@ fs.readFileSync = () => Buffer.from('Hello, ESM');
...
+>       _specifier_ as a URL.
+> 1. Otherwise, if _specifier_ starts with _"/"_, _"./"_ or _"../"_ then,
+>    1. Set _resolvedURL_ to the URL resolution of _specifier_ to _parentURL_.
+> 1. Otherwise,
+>    1. Note: _specifier_ is now a bare specifier.
+>    1. Set _resolvedURL_ the result of
+>       **PACKAGE_RESOLVE**(_specifier_, _parentURL_).
+> 1. If the file at _resolvedURL_ does not exist then,
+>    1. Throw a _Module Not Found_ error.
+> 1. Return _resolvedURL_.
```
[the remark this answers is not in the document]
guybedford: I've split this out in the latest changes. As mentioned in the discussion, this implementation or PACKAGE_MAIN_RESOLVE can easily be extended in just a few spec lines to also support what you are describing, but was just implemented here first was designed not to come with "folder mains" or "directory indexes" whatever you want to call them.

Either way, we can shift this discussion to next time now.

guybedford:
@MylesBorins thanks I've done the modules-lkgr rebase. We still need to get approval for merge on this though, so may have to wait for next meeting here.

doc/api/esm.md
```diff
@@ -1,4 +1,4 @@
-# ECMAScript Modules
+res# ECMAScript Modules
```
targos:
```suggestion
# ECMAScript Modules
```

doc/api/esm.md
```diff
@@ -146,6 +142,51 @@ fs.readFileSync = () => Buffer.from('Hello, ESM');
...
+>       **PACKAGE_RESOLVE**(_specifier_, _parentURL_).
+> 1. If the file at _resolvedURL_ does not exist then,
+>    1. Throw a _Module Not Found_ error.
+> 1. Return _resolvedURL_.
+
+**PACKAGE_RESOLVE**(_packageSpecifier_, _parentURL_)
+> 1. Assert: _packageSpecifier_ is a bare specifier.
+> 1. If _packageSpecifier_ is a Node.js builtin module then,
+>    1. Return _"node:${packageSpecifier}"_.
+> 1. Set _parentURL_ to the parent folder URL of _parentURL_.
```
jkrems: I think having this outside of the `While` makes the loop never terminate?

guybedford: Thanks!

MylesBorins:
I've rebased against modules-lkgr

MylesBorins:
CI: https://ci.nodejs.org/job/node-test-pull-request/18840/

guybedford:
I've gone ahead and squashed this PR into a single commit.

MylesBorins:
one more CI: https://ci.nodejs.org/job/node-test-pull-request/18853/

lets land if green 🎉

guybedford:
Good call - we had a long line linter failure from one of the feedback commits I've just pushed a fix for :)

guybedford:
Yet another CI: https://ci.nodejs.org/job/node-test-pull-request/18854/

guybedford:
Turned out there was another weird issue with the type checker detecting the spec usage of string interpolation as invalid types. I've switched to explicit string concatenation in the spec instead which should resolve this now finally.

https://ci.nodejs.org/job/node-test-pull-request/18855/

guybedford:
CI all green. Manually squashed and merged into modules-lkgr!

PR #28, opened 2019-01-27
Irp type implementation

guybedford:
This PR includes the specification and implementation work for:

* The IRP proposal: https://github.com/GeoffreyBooth/node-import-file-specifier-resolution-proposal
* The mode proposal (currently being considered as "type", to be discussed): https://github.com/guybedford/ecmascript-modules-mode

The implementation is exactly in sync with the specification diff here, and provided as separate commits.

### DIFF
--- a/doc/api/esm.md
+++ b/doc/api/esm.md
@@ -55,10 +55,6 @@ property:
 
 ## Notable differences between `import` and `require`
 
-### Only Support for .mjs
-
-ESM must have the `.mjs` extension.
-
 ### Mandatory file extensions
 
 You must provide a file extension when using the `import` keyword.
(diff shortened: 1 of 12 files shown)


MylesBorins:
is this trumped by https://github.com/nodejs/ecmascript-modules/pull/29?

guybedford:
This is the PR seeking consensus. #29 is the addon given dynamic modules, but we still don't know what the status of dynamic modules is right now.

GeoffreyBooth:
This corresponds with https://github.com/nodejs/modules/issues/256. If the group wants to merge in this _and_ dynamic modules, then we could merge #29 instead; but we’re assuming there’s at least a good chance that there won’t be appetite just yet for dynamic modules, so we’re hoping that this PR (without dynamic modules) can get merged in at least.

bmeck:
I will not be able to attend any meeting today since TC39 is going on, but would like to express support for this PR if there is not quorum in a meeting. If such a meeting takes place, consider myself to be a +1.

ljharb:
I'll try to attend but may not be able to for the same reason; I object to `type: 'esm'` existing (ideally even in our fork) prior to a more generic, low-level approach that allows remapping of any extension/mime to any parse goal. (one the low-level feature exists, adding a mutually exclusive `type` field as a shortcut seems fine to me)

MylesBorins:
As a note before the meeting so this isn't a surprise... If we cannot get named exports for CJS modules I am not in support of the IRP proposal.. specifically at this point I am not sure we should be shipping any interop without named exports. It appears, and I could be wrong, that if we will not be shipping interop that there is not a need for this proposal. Thoughts?

doc/api/esm.md
```diff
@@ -157,37 +149,128 @@ The resolver has the following properties:
...
+>    _Module Not Found_ error for no resolution.
+> 1. If _legacyMainURL_ does not end in _".js"_ then,
+>    1. Throw an _Unsupported File Extension_ error.
+> 1. Return _legacyMainURL_.
+
+**ESM_FORMAT(_url_, _isMain_)**
+> 1. Assert: _url_ corresponds to an existing file.
+> 1. Let _pjson_ be the result of **READ_PACKAGE_BOUNDARY**(_url_).
+> 1. If _pjson_ is **null** and _isMain_ is **true**, then
+>    1. Return _"legacy"_.
```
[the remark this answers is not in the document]
guybedford: Great question. This was originally not the plan and was added in @GeoffreyBooth's proposal here - https://github.com/GeoffreyBooth/node-import-file-specifier-resolution-proposal/commit/13669ee2f97692ee59d4d699b4a914d0f51d1740 after implementation feedback.

Basically we went ahead and implemented the default as "esm", but then when running Node's own tests I got an error on the `node tools/lint-js.js` command not working, due to there being no `package.json` inside of Node.js itself to create a CommonJS boundary, which then led to this change.

guybedford:
Thanks @targos for the review, I've integrated those changes.

GeoffreyBooth:
I think without CommonJS interoperability there’s still definitely a need for this proposal, though obviously not the parts of it that concern `import` statements of CommonJS files/packages. The proposal includes several features that are ESM-specific, like how to enable `.js` files as ESM. I would want to keep those features even without CommonJS interop, though the proposal would need to be rewritten.

doc/api/esm.md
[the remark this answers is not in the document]
targos: I see, but what's preventing us from dropping this case and falling back to file extension checking?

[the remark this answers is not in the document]
guybedford: The cases for this are actually still not set in stone at all, and this is being fleshed out in another proposal. So this may still change too.

See https://github.com/GeoffreyBooth/node-esm-entry-points-proposal - the goal was to separate them so we could merge separately and build on consensus instead of trying to get the full consensus upfront.

GeoffreyBooth:
Per @MylesBorins in the meeting today, this will be merged in Wednesday, Feb. 6 unless there are objections.

If you _have_ objections, please note them before Feb. 6 so that they can be addressed before the next meeting on Wednesday, Feb. 13.

weswigham:
@ljharb can't we agree that a mechanism should exist and that when it exists it will be mutually exclusive with this flag (and that this flag should be an alias for some kind of other more verbose configuration), but defer specifying it? Having to do that means getting into the weeds about pluggable loaders and mimes and a whole boatload of stuff that by and large doesn't matter for what we really want to solve with this flag.

GeoffreyBooth:
@ljharb I’m happy to support a `"mimes"` field or some other way of configuring file extensions to MIME types, but that feels like a separate feature that deserves its own proposal text and review. https://github.com/nodejs/modules/pull/160#issue-204604885 could be a starting point. And like @weswigham mentioned, I feel like it would be hard to design without a design for loaders.

Can we move forward without such a field for now and add it later once someone has done the work to design and implement it?

ljharb:
My concern is, what happens if we can never figure that out? Then we'd be stuck with `type` without being able to have the primitive it's sugar for.

GeoffreyBooth:
> My concern is, what happens if we can never figure that out? Then we’d be stuck with `type` without being able to have the primitive it’s sugar for.

My understanding of loaders is that they will certainly be able to control Node’s handling of file extensions, for example by sending `.ts` files through a TypeScript compiler. As loaders are very much on our roadmap, users will have a way to customize extensions and MIME types in whatever we ship.

ljharb:
I'm concerned with being able to specify it in package.json so that loaders can infer the only actually safe and reliable means of determining this: the author's intention. Otherwise, a (non-per-packaeg) loader has to unsafely make a choice for code that the person configuring the loader didn't author.

GeoffreyBooth:
Currently `"type": "esm"` conveys the author’s intention that `.js` files in that package should be treated as ESM. That’s all it does and all it has to do. I don’t see why we need a more complicated configuration block at this time. I would be happy to add one, if you’d like to design it, but I don’t see why it’s a blocker for this PR.

ljharb:
I think that privileging one remapping (`.js` means "ESM") over all the other possible ones (any extension means "CJS", "ESM", "WASM", etc) is a bad precedent to set, even temporarily.

weswigham:
> I think that privileging one remapping (.js means "ESM") over all the other possible ones (any extension means "CJS", "ESM", "WASM", etc) is a bad precedent to set, even temporarily.

I think privileging that one remapping is expressly what we're trying to do - we're expressly responding to the community's adverse reaction to not being able to have esm in `.js` files by default by making it an expressly trivial operation to, within a package, allow that. We are _intentionally_ privileging that remapping, because for many, it's the only remapping that matters. While I _agree_ that the more general system is more versatile and desirable, I do _not_ however think it's implementation should block this - especially since the exact implementation of that configuration shouldn't have any bearing on the shape or observable behavior of this flag.

ljharb:
Fair point. I'll think about that between now and the 6th, including thinking about a possible design for the more granular case.

doc/api/esm.md
```diff
@@ -157,37 +149,128 @@ The resolver has the following properties:
 
 ### Resolver Algorithm
 
-The algorithm to resolve an ES module specifier is provided through
-_ESM_RESOLVE_:
+The algorithm to load an ES module specifier is given through the
+**ESM_RESOLVE** method below. It returns the resolved URL for a
```
jdalton: Prolly should tweak any "URL" reference to "File URL".

lib/internal/modules/esm/translators.js
```diff
@@ -50,9 +50,14 @@ translators.set('esm', async (url) => {
...
 const winSepRegEx = /\//g;
-translators.set('cjs', async (url, isMain) => {
+translators.set('cjs', async function(url, isMain) {
   debug(`Translating CJSModule ${url}`);
   const pathname = internalURLModule.fileURLToPath(new URL(url));
+  const cached = this.cjsCache.get(url);
+  if (cached) {
+    this.cjsCache.delete(url);
+    return cached;
+  }
```
jdalton: ☝️ Could you explain this cache bit above.

GeoffreyBooth:
> `.node` and `.json` aren’t suddenly going to be parsed as ESM in a type esm package, i assume?

`"type"` is a description, not a configuration. `"type": "esm"` says that this is an ESM package. That’s it. What Node does with files inside ESM packages is up to Node. `"type": "esm"` doesn’t say anything about how Node should treat `.node` or `.json` files. It doesn’t even say how Node should treat `.js` files, because it’s _description,_ not configuration. _We’re_ defining Node’s default behavior for ESM packages to include that Node should treat `.js` files as ESM, but that’s not mandated by the `package.json`, as the `package.json` includes no configuration of Node. It only includes some metadata where the author is informing any consumers that this is an ESM package, whatever that means to whatever is consuming the package.

Node isn’t the only consumer of NPM packages: they’re also used by bundlers, for example. Having neutral metadata _describing_ a package is different than something like a `"babel"` field configuring how Babel should interact with a package. I’m not opposed to adding a configuration field, but that’s not what `"type"` is intended to be.

The file extension `.zip` tells all who see it that `file.zip` is a `.zip` file. It doesn’t say that you can or can’t list its contents, extract it, add a new file to the archive, etc. Some applications that are consumers of `.zip` files can do all those things; some can only extract, etc. In this sense, `.zip` is descriptive metadata, just like `"type": "esm"`.

devsnek:
what if I want mode:esm to mean .es files

GeoffreyBooth:
@ljharb @devsnek If you’d like to propose a new feature for configuring handling of file extensions, please be my guest. I don’t think the lack of such a feature is any reason to hold up this PR.

Everyone has their own personal preferences as to what the UX should look like. Completing this implementation will be a needlessly slow process if people block PRs because the UX isn’t exactly how they would prefer it to be. I ask that you please show some deference to the team that put the work into designing and coding this. We’re open to the UX changing later, as I wrote up at the top.

GeoffreyBooth:
@jdalton and @guybedford The support of `import` of CommonJS was discussed in the last meeting and there was concern there that it might not have consensus. How about we add `throw` statements wherever CommonJS is being imported via an `import` statement, but a new command line flag `--allow-commonjs-import` will bypass the exceptions? So something like:

```js
// Within the function where CommonJS is being imported
if (!process.env.NODE_OPTIONS.includes('--allow-commonjs-import')) { // Or however you check the flag
  throw new Error('Error: import statements cannot be used to import CommonJS files or packages');
}
```

@guybedford does something like that work for you? That way people can still test the CommonJS import interoperability, without necessarily “shipping” it. We can decide later whether to ship like this (with a special opt-in flag), ship with it enabled all the time (no opt-in required) or ship with it always disabled (no flag, the exception is always thrown).

@targos I believe your review comments were addressed, was there anything else? If not, do you mind dismissing your review?

@ljharb I don’t see how we could accommodate your objection even if we wanted to, as you haven’t proposed a design for your suggested new feature. I also suspect it lacks consensus, at least as a blocker to this PR, based on the comments in this thread and in other discussions about MIMEs on other threads and in meetings. Do you mind please withdrawing your objection and you can propose it as a new feature to be added in its own PR? I think that approach might find more success.

Everyone else, assuming we make the change I propose above and @targos and @ljharb are okay with dismissing their reviews, could we merge in this PR plus the `throw`/flag change on Wednesday Feb. 6?

lib/internal/modules/esm/translators.js
guybedford: Sure, basically in `--experimental-modules` we "snapshot" the `module.exports` value very carefully to ensure that we don't get any later assignments to `module.exports`. When there is no ambiguity, the `--experimental-modules` cjs loader injects all its modules directly into the ESM loader registry to provide these snapshots, even for ESM modules that haven't yet loaded to ensure this consistency. When we introduce the ability for a `.js` file to be either format, we need to separate this injection registry to ensure that a `.js` file that can be interpreted as either ESM or CJS is definitely interpreted with CJS semantics for require and ESM strict semantics for ESM, instead of whichever happened to come first.

guybedford:
@GeoffreyBooth I'd be wary updating this PR to block CJS imports, as that is a large part of the usefulness of this PR. But I can certainly get behind a flag approach if that is where consensus lies here. Let's discuss this further in the meeting to get consensus further.

ljharb:
(ftr I’ve proposed a design multiple times; a simple map of actual extension to intended extension, eg `extensions: { js: ‘mjs’ }`, but surely there’s no consensus yet)

GeoffreyBooth:
@targos and @ljharb Your flexibility is greatly appreciated. Thanks!

@ljharb We went through a consensus-seeking process to design a file extension map in https://github.com/nodejs/modules/pull/160#issue-204604885. Your suggestions were part of that discussion. We reached a consensus which I wrote up at the top of that thread. That could serve as a starting point for a new proposal.

guybedford:
Thanks all for feedback. As discussed at the meeting, let's get this merged tomorrow.

MylesBorins:
@guybedford has this been updated regarding current state of dynamic modules? I'm not in favkr of exporting only the default

guybedford:
@MylesBorins the plan was to see if we can do a follow-up to this PR with a userland approach to named exports for the coming meeting as an alternative to #29. It would be easier if we can iterate as we go I think.

MylesBorins:
Also worth mentioning there is a massive conflict with upstream we should make sure is resolved before landing anything

doc/api/esm.md
```diff
@@ -18,14 +18,10 @@ interoperability support, specifier resolution, and default behavior.
...
-The `--experimental-modules` flag can be used to enable features for loading
-ESM modules.
-
-Once this has been set, files ending with `.mjs` will be able to be loaded
-as ES Modules.
+Modules are enabled by default for files ending with `.mjs`:
 
 ```sh
-node --experimental-modules my-app.mjs
+node my-app.mjs
```
MylesBorins: why has the flag been removed? If we upstream this we will still require a flag for the feature

GeoffreyBooth:
> i will not object to this landing in its current form, but there will need to be more consensus building and work to make this something that can ship upstream.

This is hardly meant to be the end of the work on our new implementation. This is just one part of Phase 2-3. There’s plenty more to come and the stuff in here is all subject to change in future PRs.

doc/api/esm.md
GeoffreyBooth: Is it removed in the code itself? If not, can we fix the doc in a follow-up PR?

MylesBorins:
> This is hardly meant to be the end of the work

💯. Not intending to sound snarky but I want to have my opinion on the record.

MylesBorins:
opened PR for rebase --> https://github.com/nodejs/ecmascript-modules/pull/30

thankfully this lands cleanly on top of it 🎉.

doc/api/esm.md
guybedford: Thanks for spotting that, that was a leftover doc change from the demo implementation which I've corrected now. This PR hasn't removed `--experimental-modules` at all.

guybedford:
Thanks @MylesBorins for the review - your points are heard. A huge thank you for fixing up the rebase as well, that did not sound fun.

jdalton:
Closing since #30, was merged.

GeoffreyBooth:
@jdalton Isn’t this PR meant to merge in after #30? That was just the rebase from upstream, and this is the new additions for our new implementation?

jdalton:
Ah, by bad. Sorry.

GeoffreyBooth:
@MylesBorins is this ready to merge? Do you mind cutting a new release afterward?

guybedford:
Merged into lkgr. Thanks all for feedback.

PR #29, opened 2019-01-27
WIP [Do not merge] - Irp type dynamic modules

guybedford:
This branch implements the Dynamic Modules v8 implementation and then builds it into the Node.js usage as DynamicModuleWrap alongside ModuleWrap, which can be used to construct dynamic modules for CJS.

The implementation uses the approach `{ ...exports, default: exports }` as the dynamic module shell for CJS and core modules.

So that this branch is effectively identical to #28, but with `import { name } from 'cjs'` support :)


ljharb:
Note, though, that the pushback wasn't about this approach not being feasible, it was that perhaps the parts of the spec should be *remove* such that it would allow node to do this on its own, without TC39 buy-in.

jdalton:
@ljharb 

There was push back on multiple fronts.
@guybedford's [comment here](https://github.com/nodejs/ecmascript-modules/pull/31#issuecomment-462158587) is a nice summary.

ljharb:
Perhaps i missed some of them; it’d be great to get a summary. From what i saw, especially given the previous consensus/commitment to accommodate node, the objections were overcomeable.

jdalton:
The pushback I remember combined with notes:
  * Source Text Modification
    * Some didn't want any modifications (to source text module or abstract module records)
    * Others didn't like Node-only branching
    * Others were concerned with weakening the static guarantees
  * Throwing on `export * from`
    * Preferred `default`-only export to named exports with throwing
    * Some worried it would cause devs to avoid `export * from` altogether

PR #34, opened 2019-02-16
Modules lkgr rebase

MylesBorins:
Rebase had conflicts. No more conflicts

PR #35, opened 2019-02-16
Modules lkgr rebase no loader

MylesBorins:
This is the same as https://github.com/nodejs/ecmascript-modules/pull/34 but without 531b6d4

as discussed in https://github.com/nodejs/modules/issues/259

/cc @jkrems


MylesBorins:
CI: https://ci.nodejs.org/job/node-test-pull-request/20817/

MylesBorins:
@jdalton and @jkrems you two were ones who I know had opinions about this. Can you please approve to make it clear you support this

doc/api/cli.md
```diff
@@ -499,6 +492,18 @@ added: v2.4.0
 
 Track heap object allocations for heap snapshots.
 
+### `-m`, `--type=type`
+
+When using `--experimental-modules`, this informs the module resolution type
+to interpret the top-level entry into Node.js.
+
+Works with stdin, `--eval`, `--print` as well as standard execution.
```
jkrems: (Not sure if in scope for this review.) It may be good to explicitly note/warn that it won't work for CLI tools via shebang.

doc/api/esm.md
```diff
@@ -132,127 +138,155 @@ fs.readFileSync = () => Buffer.from('Hello, ESM');
...
+>       1. Return the result of **PACKAGE_MAIN_RESOLVE**(_packageURL_,
+>          _pjson_).
+>    1. Otherwise,
+>       1. Return the URL resolution of _packageSubpath_ in _packageURL_.
+> 1. Throw a _Module Not Found_ error.
+
+PACKAGE_MAIN_RESOLVE(_packageURL_, _pjson_)
+> 1. If _pjson_ is **null**, then
+>    1. Throw a _Module Not Found_ error.
+> 1. If _pjson.main_ is a String, then
```
jkrems: This means that we create an entrypoint field that is closed for extension ("we can never add a new extension w/o breaking the else: CJS") unless it's paired with a 2nd `package.json` field (`"type": "module"`). This feels a bit awkward. I assume reusing `main` is temporary?

GeoffreyBooth: Yes, it was discussed in the PR where this was landed that I was going to propose changing this, as the current implementation makes dual CommonJS/ESM packages impossible. That will require new `package.json` configuration, though, so it’ll take a new proposal and more review etc., so we allowed merging this in with the understanding that dual-type packages would be addressed in the future.

doc/api/cli.md
GeoffreyBooth: I think this is just a rebase? Could you open a PR like https://github.com/nodejs/ecmascript-modules/pull/33 to change this, or an issue on this repo to remind us to address this?

jkrems: Done: https://github.com/nodejs/ecmascript-modules/pull/37

doc/api/esm.md
jkrems: Added https://github.com/nodejs/ecmascript-modules/issues/38 to also cover future extensions. For me that's even more important than dual-mode packages because it's not limited to a transitional time.

MylesBorins:
landed in bc72470788...968c8f1ff9

PR #41, opened 2019-02-26
Exports main

guybedford:
This supports the package.json "exports" main to allow dual mode workflows, as described in https://github.com/nodejs/modules/issues/273.

In addition some minor refinements around file extension handling are provided as well, specifically disabling support for `.node` and `.json` mains in CommonJS packages when loaded from ESM.


MylesBorins:
@guybedford can you update the commits to have a namespace e.g. `esm:`

GeoffreyBooth:
> Per comment on the issue; without extension lookup i don’t think we should merge this.

Per the issue discussion, this PR doesn’t involve extension lookup. Even if extension lookup is added, this PR would still apply.

ljharb:
I disagree; with extension lookup, there's no need for an additional field at all, since the file extension would determine parsing goal (and a separate field to override parse goal per extension would continue to be useful).

GeoffreyBooth:
> I disagree; with extension lookup, there’s no need for an additional field at all, since the file extension would determine parsing goal (and a separate field to override parse goal per extension would continue to be useful).

As was noted in that thread, a single field only works if the two files are side-by-side differing only in extension, e.g. `./index.mjs`/`./index.js`. A very common, if not predominant, developer workflow is to have ESM code in a folder like `src` and the transpiled CommonJS code in a separate folder like `dist`. Therefore there would be separate entry points like `./src/index.mjs` and `./dist/index.js` that are impossible to reference in the extension-searching approach. Note also that this point still applies even if one uses separate extensions for their ESM and CommonJS code.

ljharb:
@GeoffreyBooth in that model, you'd transpile/copy two files, and have `dist/index.mjs` and `dist/index.js`.

GeoffreyBooth:
> @GeoffreyBooth in that model, you’d transpile/copy two files, and have `dist/index.mjs` and `dist/index.js`.

That’s a much bigger burden on developers, to have essentially two build pipelines. Also there are a lot of developers who will want to use `.js` for both ESM and CommonJS files, which isn’t possible if they need to share the same folder.

About the only advantage `"main": "./index"` offers is that it doesn’t involve adding a new field to `package.json`. Conversely, adding a new field offers many advantages:
- Users can use `.js` for everything if they want.
- Users can have separate entry points in separate folders, without needing unnecessary build steps like copying files or rebuilding already-ESM files or creating symlinks. The `package.json` fields could point directly to the real files without any shenanigans necessary.
- An `"entrypoint"` field that requires a full extension, regardless of whether or not extension searching is enabled for `import` statements, provides a way for build tools and other environments to find the ESM entry point for a package without needing to implement Node’s extension searching algorithm (which might very well be impossible once loaders are added).

For all these reasons and a few I can’t think of at the moment, we judge `"exports"` to be a better solution than `"main": "./index"`. Yes it involves creating a new field, but there are big advantages to doing so.

ljharb:
It's a fair point that even with extension lookup, we might still want a way to have an alternative "main". However, since deep imports are important as well, imo it's not really sufficient unless we can provide an alternative *root directory* for the package, ie, `"esm-root": "src"` or something.

Re `* Users can use `.js` for everything if they want.`, I don't think this is a benefit unless you can replace `.js` with "any extension they like", and unless you do replace "everything" with "ESM", since ESM JS is not "everything".

(Another alternative is to not use a `dist` dir, but instead, to write `.mjs` files in `src` and use a build process to produce the `.js` ones.)

Every build tool already implements node's searching algorithm - in practice this has proven to not be a hardship. Similarly, loaders would need to expose a way for build tools to tap into their resolution algorithm, otherwise they wouldn't work for build tools anyways (unrelated to extension searching).

GeoffreyBooth:
@ljharb As you know, one of the goals of the import file specifier resolution proposal was to support ESM in `.js` files. The _reason_ to support ESM in `.js` files is so that users don’t have to work with multiple file extensions representing JavaScript—or to put it more bluntly, to not need to use `.mjs`. We know from user feedback that many users strongly dislike `.mjs`, and therefore we’re trying to provide an ESM implementation that doesn’t require the use of `.mjs`. Users are welcome to _opt in_ to using `.mjs`, but whenever possible they should not be _required_ to use it.

The `"main": "./index"` approach requires the use of either `.mjs` or `.cjs` (or some other secondary extension, if mapping becomes possible) for some of the users’ JavaScript files. Many users will find this unacceptable. I am one of them.

`"exports"` provides a way for defining separate CommonJS and ESM entry points without requiring users to use either `.mjs` or `.cjs`. Again, people still _can_ use `.mjs` if they desire, but they would be able to create dual-CommonJS/ESM packages without needing to use `.mjs`. It also corresponds with what users are already used to, where they define the CommonJS entry point in `"main"` and the ESM entry point in `"module"`. It’s not much of a stretch to switch from using `"module"` to using `"exports"`.

I support a way to map extensions, whether via some new `package.json` configuration or via loaders; but I would ask that you please stop blocking PRs where the intent is to provide a solution that doesn’t _require_ the use of `.mjs` or some other secondary extension for some of users’ JavaScript. We’re trying to build an implementation where users can keep using `.js` if they prefer. Even in this new implementation there’s plenty of encouragement for them to use `.mjs`: better performance, and no need for `--type` or `"type"`. But if users still want to keep using `.js` for all their files, whether CommonJS or ESM, we should let them.

ljharb:
I agree they shouldn't be required to use `.mjs`, but no, it should *not* be opt in - it should be opt *out*, ie, the default (which it is). It's totally fine to offer a way to opt out of `.mjs` into another extension; but just because we've only heard about one desire (to parse `.js` as ESM) doesn't mean that one combination out of all the others should be privileged.

Your desire - to avoid using `.mjs` - is totally valid and something we should provide a way to opt *in* to. However, this can and should be done in a generic way, and "exports" does not provide this.

I'm going to continue blocking PRs that I don't think are a good fit - half measures are worse than none imo. For me, the goal of "let people not use `.mjs`" or "let people parse `.js` as ESM" is one I'm happy to provide, but I find to be one of the *least* important goals we have. Any effort that prioritizes that goal over others isn't going to get support from me, and it's not reasonable to ask me to sacrifice my priorities just because you disagree with them.

devsnek:
not entirely in the same context but this addresses the points of mjs because of course mjs has to keep coming up again and again :( please read this: https://github.com/nodejs/modules/issues/253#issuecomment-457056757

i see absolutely no reason for node to encourage any file extension for new stmr files besides `.mjs`. handling legacy cases is important but legacy is legacy, not the best design nor the design we should encourage.

GeoffreyBooth:
> I agree they shouldn’t be required to use `.mjs`, but no, it should _not_ be opt in - it should be opt _out_, ie, the default (which it is).

`.mjs` is already the default in this implementation. You opt out by using `"type": "module"` or `--type=module`. If you want to provide _another_ way to opt out, that’s explicitly file extension-based, that’s fine; but that’s a new feature, that you should propose.

Since we both agree that users shouldn’t be required to use `.mjs`, and since it’s clear that the `"main": "./index"` approach requires the use of `.mjs` (or `.cjs`), can you drop your objection to this PR? Unless you can think of another way to support dual-CommonJS/ESM packages, it would seem that this is the best option on the table. We can always replace this with another implementation later if someone comes up with one. Perhaps when the file extensions map is proposed, part of that proposal can be a revision of how dual packages can work once extensions can be mapped.

ljharb:
It has been proposed - every github comment I’ve made about it is a proposal.

I continue to think that we should not proceed with this without first figuring out extensions lookup and mapping.

GeoffreyBooth:
@ljharb Can we agree that if/when automatic extension resolution lands, part of that PR can involve changing this? Because if automatic extension resolution ends up not shipping as part of upstream, I’d still rather support dual CommonJS/ESM packages somehow rather than letting this feature get blocked as well.

ljharb:
@GeoffreyBooth without a form of dual support, we'll never get consensus to unflag anyways, so there's no value in shipping a piece now with the intention to change it later.

GeoffreyBooth:
Let’s boil this down. Even if we get automatic extension resolution, I will still object to `"main": "./index"` for the many non-automatic extension resolution reasons stated above: it doesn’t support entry points in separate folders, and it doesn’t support users who want to use `.js` for both their ESM and their CommonJS JavaScript files. It’s way more confusing, IMO, than following what users are used to with the separate `"module"` field defining the ESM entry point. `"main": "./index"` also doesn’t scale _at all,_ like for browser entry points and the like, whereas creating an `"exports"` key that explicitly accepts an object in its verbose form (to be specified later) creates an opening for future functionality along those lines, as well as mapping paths.

So if you want to propose `"main": "./index"`, even though you won’t open an issue for it, then mark me as blocking your proposal. You can continue to block this one and we’re at a stalemate and so dual CommonJS/ESM won’t ship as part of the upstream PR. I don’t know why you keep tying this to automatic extension resolution, unless you’re trying to force that upon the group by holding this feature hostage to it? But I reject that one requires the other. And even if we ship automatic extension resolution, I still reject `"main": "./index"`. So you can either stop blocking this, come up with another idea that can achieve consensus, or I’m content with leaving this as a stalemate or going to a vote.

ljharb:
I'm not sure what you're objecting to there; even *omitting* a main in CJS *already* looks for `index` using extension resolution. That certainly needn't be a *requirement*, but I don't understand why you'd object to people continuing to do that with ESM. What does it even mean to "reject main: /index"??

A separate mechanism for what you're talking about makes sense, sure, but *on top* of extension resolution.

GeoffreyBooth:
> I’m not sure what you’re objecting to there

Why are you objecting to this PR? I thought it was because you wanted automatic extension resolution so that the two entry points can both be defined via `"main"`, where the value gets resolved with different extensions for CommonJS and ESM. So `"main": "./index"` where there’s an `index.js` and an `index.mjs`, and CommonJS gets the former and ESM gets the latter. You wrote above that you wanted to use this `"main": "./index"` approach because you didn’t want to add a new field to `package.json`. Is this no longer your preference? If not, how _do_ you want to support dual ESM/CommonJS packages?

I want to add a new field `"exports"` to define the ESM entry point. Right now in this implementation it needs to be a full filename including extension, but if we add extension searching to the implementation overall then we can consider also adding extension searching to `"exports"`. Would that be acceptable to you? If you’re okay with a separate field `"exports"` for the ESM entry point, why can’t we add that now and make it support extensionless filenames whenever we bring extension searching into the implementation generally?

ljharb:
Yes, that's the approach I want to use - but nothing about it forces *you* to use it.

What you want also seems like a reasonable problem to solve (but I think a mechanism to solve it should also handle deep imports, like `moduleRoot` or something) - but I think that's a secondary priority both in importance and in ordering.

GeoffreyBooth:
Why are you blocking this PR and what needs to happen to get you to release your block? Be specific. I’m tired of this back and forth.

ljharb:
I've described it already, many times, upthread and on other threads. See https://github.com/nodejs/modules/issues/283 for the proposal you asked for.

GeoffreyBooth:
> I’ve described it already, many times, upthread and on other threads. See [nodejs/modules#283](https://github.com/nodejs/modules/issues/283) for the proposal you asked for.

This doesn’t answer my question. Are you saying that if the proposal in https://github.com/nodejs/modules/issues/283 is implemented, you’ll drop your objection? That’s it?

I don’t see what a file extensions map has to do with this. Why is the existence of a file extensions map necessary for `"exports"`? It would seem not to effect this PR.

And what if your file extensions map doesn’t achieve consensus? You’re going to keep blocking this PR regardless?

ljharb:
Sorry for being unclear; this proposal imo is blocked on extension resolution; the "type"/"mode" field I'm blocking on unflagging pending the extensions map.

PRs don't matter (to a flagged/experimental implementation); decisions do. If a decision is made in either direction on something, we can certainly reevaluate any existing objections and PRs.

GeoffreyBooth:
> the “type”/”mode” field I’m blocking on unflagging pending the extensions map.

I have no idea what you’re talking about. The `"type"` field won’t be “unflagged” until all of `--experimental-modules` is unflagged. It’s also been part of the implementation for weeks, and several PRs have been built on top of it. It’s been repeatedly approved by consensus, including today’s approval of a PR regarding error messages on mismatches between `"type"` and `--type`. If you want to propose reverting all those consensus approvals, you’re welcome to try, but what does objecting to that have to do with this PR?

This is the last time I’m going to ask: exactly what would it take for you to drop your objection to this PR? Please be specific and clear. It’s apparently not just that you want the group to approve https://github.com/nodejs/modules/issues/283. What else are you demanding?

ljharb:
I explicitly gave my consensus for "type" with the caveat that it could not be unflagged without some form of an extensions map.

For this PR, my answer remains the same as https://github.com/nodejs/ecmascript-modules/pull/41#issuecomment-467866994.

I'm going to avoid responding further in this thread for now, since we seem to be going in circles where you ask me questions i've already answered. My block remains.

GeoffreyBooth:
@mylesborins please add this PR to the next agenda. I would like a vote. This block is not germane to this PR nor is it requesting something that has consensus. @ljharb's demands are impossible to accommodate.

GeoffreyBooth:
Aside from the above, does anyone else have concerns with this PR or with dual CommonJS/ESM package support in general? I think there were some concerns mentioned in today's meeting but they flew by too quickly for me to note. Please mention them here, I'd like to get them addressed soon so we can support dual packages either before upstreaming or shortly afterward.

Alhadis:
As long as named/Node built-in modules are still loading perfectly fine for dynamic imports, I've got no reason to crap my pants.

~~~js
const {readFileSync} = await import("fs");
const {spawn, exec} = await import("child_process");
~~~

MylesBorins:
@Alhadis that comment is a bit out of scope. If you do a dynamic import it will return a promise that resolves the namespace of the object. If you are in an async function you can await the result and destructure it, which looks similar but is not named exports.

This is all behavior defined by the language though, not Node.js

devsnek:
@GeoffreyBooth I have concerns about the existence of dual mode. multiple sources for a single specifier is a huge and dangerous footgun.

targos:
@devsnek When ESM are out in a release, it seems like dual mode will be the easiest way for a library to support new and older versions of Node.js.

devsnek:
@targos assumedly your library can just be loaded as cjs on all platforms. as with any new features from the last ten years, if you want back compat you have to restrict what you use.

maybe there is some other solution, but having two separate versions at runtime is very broken and not an apparent bug when you have to fix it.

targos:
I was more thinking about two separate entry points into the same library to provide named exports.

devsnek:
@targos that may work but i would strongly disagree that it is the best way to do so. if the goal is adding named exports we could add a field to package.json and get rid of the dual mode risk. or who knows, one of the dynamic modules things might actually happen.

GeoffreyBooth:
@devsnek Doesn't the current `--experimental-modules` support dual-mode packages? How does it deal with the problem of a single specifier potentially pointing to separate files in CommonJS vs ESM?

devsnek:
@GeoffreyBooth it's a problem for any formal the esm loader supports that cjs doesn't. the fix is to make cjs either accept those formats or throw.

GeoffreyBooth:
> the fix is to make cjs either accept those formats or throw.

So I understand you, is this how the problem is solved in the shipped implementation? A `require` of a dual specifier loads the ESM version into CommonJS? Or does it throw on the CommonJS side?

devsnek:
@GeoffreyBooth it isn't currently fixed. i believe the correct fix would be to assign `module._extensions['.mjs']` before any of the other extensions.

GeoffreyBooth:
> @GeoffreyBooth it isn’t currently fixed. i believe the correct fix would be to assign `module._extensions['.mjs']` before any of the other extensions.

Can you please explain this a bit more? Are you referring to the CommonJS loader?

devsnek:
@GeoffreyBooth right now the cjs loader will find `.js` before it finds `.mjs` and successfully load the file, sometimes creating the situation known as dual mode. if it searches for `.mjs` first, it will throw when it finds it and prevent such a situation.

weswigham:
Just to be clear, doing what @devsnek suggests would force packages into a single mode in post-esm node, but still allow them to ship a downleveled version of their package side-by-side for older versions of node (since they'll ignore the `mjs` file and go straight for the `js`), assuming extension resolution on `main`. IMO, this scheme makes the most sense to me, since then you don't end up with two instances of the same package knocking about in post-esm node, and you still support the ability to provide a cjs-compatible version of your package under the same name for older versions of node.

devsnek:
@weswigham unfortunately its not perfect 😢. lets say lodash ships esm and cjs source like above. in a version of node with my proposed change, you wouldn't be able to load lodash via require.

weswigham:
> you wouldn't be able to load lodash via require.

Unless, ofc, we create some way to load `esm` via `require`, which, IMO, would be ideal, since without such a feature upgrading a package to `esm` is always a breaking change for all existing consumers. :(

MylesBorins:
@weswigham we had dug into requiring ESM quite a bit... the only solution that was "technically" possible was to return a promise... and at that point you may as well just use dynamic import.

weswigham:
That was just without "ugly" things like turning the event loop during the `require` call to let the module resolve, right? :3

devsnek:
@weswigham i've dipped my toes in that solution a bit (https://github.com/devsnek/syncify) and the tl;dr is that running parts of the event loop without others subtly breaks a lot of things resulting in catastrophic failures. this is mostly due to people assuming the order of the event loop but there's not much we can do about that :(

weswigham:
Hmm... What if we created a mini-event loop just for esm resolution activities and only turn that (up till the execution stage), while keeping the main event loop paused? This way nothing else queued by prior cjs execution is executed (yet) but esm still resolves.

weswigham:
@devsnek I know _very_ little about how the node internals actually work, but [this](https://github.com/weswigham/ecmascript-modules/commit/b1d5ee65390bf85d4b798c71a2fa91e9051f4c3f) aligns with what I was thinking of. Is there some bad consequence to using event loops in this way?

GeoffreyBooth:
@devsnek There was a long discussion about dual sources for the same specifier [here](https://github.com/GeoffreyBooth/node-import-file-specifier-resolution-proposal/issues/1) between @robpalme and @Fishrock123 and @bmeck. It resulted in the [“Dual instantiation” section of the IRP proposal.

devsnek:
@weswigham there's no guarantee that everything queued by your function call in the first function is handled by the time you put the loop back. the second function has the same problems I discussed earlier.

@GeoffreyBooth having two entry points in package.json is a literal invitation into dual mode, something everyone agrees is bad. I respect the existence of dual mode but we cannot be encouraging people to shoot themselves in the foot. any system where cjs and esm have separate entrypoints is a non-starter to me.

ljharb:
Everyone does not agree dual entry points is bad, ftr.

devsnek:
@ljharb apologies, I was speaking from the context of the issue Geoffrey linked me, not the conversation here. should'a made it more clear

GeoffreyBooth:
> @GeoffreyBooth having two entry points in package.json is a literal invitation into dual mode, something everyone agrees is bad. I respect the existence of dual mode but we cannot be encouraging people to shoot themselves in the foot. any system where cjs and esm have separate entrypoints is a non-starter to me.

I would appreciate it if we could tone down the rhetoric a bit. We’re just trying to solve a technical problem here. There’s no moral outrage happening.

I’m trying to reconcile “I respect the existence of dual mode” with “any system where cjs and esm have separate entrypoints is a non-starter” and I’m confused. Even in the implementation as it stands, one can create a dual-ESM/CommonJS package if one is willing to use deep imports. To take the `src`/`dist` example from above, a dual package could take the form of `import 'foo/src/index.mjs'` and `require('foo/dist/index.js')`. The ESM entry point doesn’t need to be specified in `package.json` at all; the reason we want it defined there is so that the bare specifier (`import 'foo'` or `require('foo')`) can be used instead of the full path to the filename.

And the `"main": "./index"` approach where there’s an `index.js` and an `index.mjs` side-by-side is still defining two separate entry points; they just happen to differ only in file extension. The advantage of `"exports"` is to allow the two entry points to differ more than just the extension, for example to be in separate subfolders.

Any system where one can `require` and `import` the same specifier, like `'foo'`, has the potential to be a footgun, as was discussed in depth in that other thread. But there’s really no reason for users to want to do so, so it shouldn’t come up often (if at all). Whereas the consequence of disallowing `require` and `import` for the same specifier would be to make bare specifiers of dual packages impossible, which will greatly hinder adoption of ESM in NPM packages.

devsnek:
@GeoffreyBooth i don't want to make it easy for dual mode to occur. this addition literally encourages people to use dual mode. node shouldn't make it easy to get into tricky situations.

> `import 'foo/src/index.mjs'` and `require('foo/dist/index.js')`

these are two separate specifiers, there's no problem here.

> Whereas the consequence of disallowing require and import for the same specifier would be to make bare specifiers of dual packages impossible, which will greatly hinder adoption of ESM in NPM packages.

i completely agree this is an issue, i just don't consider this pr an acceptable solution. I'm playing locally with my own various changes but i haven't hit anything nice yet.

my current stance is therefore that backwards compatibility would need to function like any other new syntax, where you have to avoid using it. i am hopeful we can come up with something that means this won't be the case, but i don't agree that multiple entrypoints is better than just continuing to ship cjs.

GeoffreyBooth:
> i completely agree this is an issue, i just don’t consider this pr an acceptable solution. I’m playing locally with my own various changes but i haven’t hit anything nice yet.

This is in Phase 3, so there’s no hurry to merge it in; but I want to get some solution for dual CommonJS/ESM in sooner than later. The longer `--experimental-modules` sits out there without a way for users to achieve dual packages, the more we’re sending a signal that dual packages might not be supported in Node. We don’t want the NPM registry to fill up with duplicative packages like `lodash` / `lodash-es` while we work on this.

If you can come up with a better solution, by all means please propose it. I acknowledge that this one has its own issues, but so far I don’t know of anything better. I’m also totally fine with merging this in and replacing this with a better solution later.

> my current stance is therefore that backwards compatibility would need to function like any other new syntax, where you have to avoid using it. i am hopeful we can come up with something that means this won’t be the case, but i don’t agree that multiple entrypoints is better than just continuing to ship cjs.

I don’t understand this.

weswigham:
> there's no guarantee that everything queued by your function call in the first function is handled by the time you put the loop back. the second function has the same problems I discussed earlier

I could change the first function to pass `Sync` as a parameter to the callback and have it automatically run the secondary event loop to completion before closing it (rather than relying on all work being `sync`'d as now). That shouldn't have either issue.

devsnek:
@weswigham i'm not 100% sure but changing the event loop may also cause existing handles to crash as i believe handle specific data is sometimes associated with the loop object.

@GeoffreyBooth 

> I don’t understand this.

tl;dr i think encouraging people to only ship cjs is better/safer/etc than encouraging them to ship something akin to dual mode.

weswigham:
> i'm not 100% sure but changing the event loop may also cause existing handles to crash as i believe handle specific data is sometimes associated with the loop object

There a way to test this? AFAIK by calling `uv_stop` I'm preventing the original event loop from turning until the call stack returns back to node's core run loop function that calls `uv_run` on it again (by which point I've restored the loop to it's original state) - the goal is that none of the original handles are executed while the secondary loop is active, so none of them should be able to witness the altered state. (And since we're spinning the secondary loop synchronously this is pretty easy for us to do, I think)

devsnek:
@weswigham I'm hopeful that is the case, I'm just concerned about unexpected side effects. if your experiments work we might be able to dodge dual mode.

could you drop by #node-dev and/or #libuv on freenode to discuss this further?

weswigham:
> sidenote: it's also impossible for a promise to resolve to another promise so you don't need to call Await recursively

Oh, good. I was worried I needed to unwrap recursively.

 > and you're not advancing node's tick queue, which may result in promises tied to events in that to never resolve.

Is that the `tick_callback` on the environment? Is that meant to execute before/after microtasks/main loop tasks? Actually, what can register a tick callback? It looks like it's setup by `process` and is what runs the microtasks queue while also tracking unhandled rejections; I think we'd actually want a separate tick callback (and, for that matter, a secondary microtasks queue since it doesn't seem tied to the event loop like I thought, but rather the isolate heap) for the secondary loop.

devsnek:
@weswigham i think we should discuss this elsewhere as its going off the topic of this pr.

GeoffreyBooth:
> tl;dr i think encouraging people to only ship cjs is better/safer/etc than encouraging them to ship something akin to dual mode.

I think the group is in agreement that people need to be able to ship ESM via NPM packages, and those same packages need to include CommonJS during a somewhat long transition period. So we need to support that somehow, whether you call it dual mode or something else. I don’t think encouraging people to not publish ESM to the NPM registry is an option.

What exactly is your concern? As in, what would need to be resolved in order for you to feel comfortable with recommending people publish a package to the NPM registry that can be both `require`d and `import`ed? Is it that you want to make sure an error is thrown somehow if someone `require`s and `import`s the same package into the same project? (Per [this other thread](https://github.com/GeoffreyBooth/node-import-file-specifier-resolution-proposal/issues/1) that doesn’t appear to be possible in a determinative way.) Is it something else? Can you please explain your concern specifically enough that someone besides you can attempt to address it?

If the concern is the fact that the same specifier can be both `require`d and `import`ed, like `require('foo')` and `import 'foo'`, and you’d like to prevent that, how do you propose to address that issue if extension searching is added to the implementation? Without extension searching, the only possible double instantiation is for a bare specifier (`'foo'`); but with automatic extension resolution, `require('./foo')` could resolve to `foo.js` and `import './foo'` could resolve to `foo.mjs` and therefore you have the same `'./foo'` specifier resolving to different files in CommonJS and ESM, just as `'foo'` could resolve to different things for a package with two entry points (e.g. `src/index.mjs` and `dist/index.js`). I would think that if double instantiation of `'./foo'` is acceptable, then double instantiation of `'foo'` should be as well.

weswigham:
> If the concern is the fact that the same specifier can be both required and imported, like require('foo') and import 'foo', and you’d like to prevent that, how do you propose to address that issue if extension searching is added to the implementation?

That's why I've been talking about how we can make `require('esm')` work - if we do that, then require'ing a specifier and importing them can return the same namespace.

devsnek:
@GeoffreyBooth 

> I don’t think encouraging people to not publish ESM to the NPM registry is an option.

in such a situation, personally i would encourage people to dump cjs (and older versions of node) entirely, and only publish esm. Some people wouldn't want to drop support for older node, so they would stick to cjs.

> Can you please explain your concern specifically enough that someone besides you can attempt to address it?

my concern is that `require('x').y === (await import('x')).default.y` will be really easy to accidentally make false, resulting in difficult bugs. big packages like graphql have had this exact issue when trying out the existing esm implementation, and i don't think actively encouraging the situation with two entry points in package.json is helping.

@weswigham and @addaleax and i had a chat on irc earlier and determined that (with some invariants, and modifications to node core) it should be possible to `require('esm')` (@bmeck it turns out that moving loaders out of the same thread actually helps a lot here 😄)

assuming that `require('esm')` can be finalized, that is my preferred alternative. packages with both cjs and esm source would then only load the esm source on new node and only load the cjs source on old node. (you could at that point still propose multiple fields in package.json, although i still have some design reservations about that)

GeoffreyBooth:
> @devsnek assuming that `require('esm')` can be finalized, that is my preferred alternative. packages with both cjs and esm source would then only load the esm source on new node and only load the cjs source on old node.

I publish the `coffeescript` package on NPM, and I need to support back to Node 6. So I’ll be publishing CommonJS source for a while. I also want to publish an ESM version at the same package name, so that Node 12+ users can do `import { compile } from 'coffeescript'`. That’s my use case, and I think it’s pretty common for any publisher of any public NPM package.

It’s not an option for me to publish the ESM version under a different package name, e.g. `coffeescript-es`, nor can I release a major version bump (going from CoffeeScript 1 to 2 was a huge thing, and if I bump to 3 people will assume there are huge breaking changes in the language). So I really just need to add ESM support as a semver minor version bump, with the CommonJS version continuing to be exported like it is now.

If we take the approach that when `coffeescript` becomes a dual package, `require('coffeescript')` in Node 12 pulls the ESM version, then we have the weird result where `require('coffeescript')` in Node <=11 uses my package’s CommonJS sources and in Node 12+ it uses my package’s ESM sources. This seems like it has the strong potential for problems.

devsnek:
@GeoffreyBooth what you want to do would work with require(esm). I'm trying to enable your use case here. we're in agreement.

MylesBorins:
I was under the impression that `require('esm')` had zebra striping issues if we wanted to create a synchronous API. Was your idea something that returned a promise or was transparent?

weswigham:
Transparent. I don't see how there would be zebra striping issues, since the esm graph you `require` presents it's namespace before it executes.

guybedford:
@weswigham zebra striping refers to an approach to deal with circular references between ESM and CJS in such a context, based on combined graph analysis. It may well be possible to implement it within this approach. But I'd be more concerned about the upgrade path issue and top-level await support in future now that that is making progress again.

GeoffreyBooth:
> what you want to do would work with require(esm).

This assumes that the ESM code and the CommonJS code are perfectly equivalent. They might not be. Any bugs in how Babel or whatever tool transpiles ESM into CommonJS would mean that my package behaves differently when required in Node 12 versus Node 11.

devsnek:
@GeoffreyBooth i don't think bugs in your transpiler should be within the scope of this group's concerns. we don't even know if people are using transpilers.

weswigham:
TLA support is nonproblematic. If it's a nonblocking variant (which is the least problematic, IMO), just returning the namespace and letting it initialize async is great. If it's a 'blocking' variant, TLA within a module can be detected and we can return a promise to the namespace instead. (Or not - depending on how you see it, the blocking variants are about sequencing within the esm graph, not without. In any case, both can be supported) But both of those actually only affect evaluation and we _do_ have a namespace available after only instantiation, so unless something drastically changes there, I don't see an issue. Zebra striping is all about preserving execution order of dependencies and having a namespace available for a module while it's still executing, both of which at least my idea for `require('esm')` keeps handily.

weswigham:
@geoffreybooth
> Any bugs in how Babel or whatever tool transpiles ESM into CommonJS would mean that my package behaves differently when required in Node 12 versus Node 11.

Babel historically is very willing to issue updates to fix this kind of issue. (Rather: they've broken module transform users twice in the past to match node's "expected" semantics, so they do seem to care) Also would be just as big an issue, if not bigger, with dual mode packages (since they'd still be executing cjs in all versions of node), so I don't really see your point.

GeoffreyBooth:
> i don't think bugs in your transpiler should be within the scope of this group's concerns. we don't even know if people are using transpilers.

The point is that `require('coffeescript')` would _load different code_ in Node 12 versus Node 11. That could be a cause of bugs.

devsnek:
@GeoffreyBooth that's a semver major upgrade, people should already expect every module they have installed to break.

aside from that, if your module is shipping untested sources it seems like your module's fault, not node's fault.

weswigham:
> the upgrade path issue

Is the biggest reason I want `require('esm')` to work~

With it, and named exports for cjs, a library need only migrate it's API off of export assignments of non-namespacy things, and then it can (mostly) transparently swap to esm under the hood (just removing the `default` fallback import - and if we provide a way to suppress that, like the `__esModule` marker, then it really _is_ transparent) without it's public API changing.

GeoffreyBooth:
@devsnek Can you please open an issue on the main repo explaining your proposal?

PR #46, opened 2019-02-28
esm: experimental wasm modules

MylesBorins:
This is a first pass at a loader for wasm modules.

It offers support for both named imports as well as deafult.
It may be possible to move some of this logic down to C++, but
I think the current implementation might be good enough. It doesn't
currently use the ESM cache, but AFAICT it isn't currently exposed.

Will dig in more with tests / examples / cache after initial
implementation gets feedback.

/cc @linclark @lukewagner @tschneidereit @sunfishcode


devsnek:
related: https://github.com/nodejs/node/pull/18972

i think we should wait for wasm modules to be specified, at which point v8 would provide them for us.

lib/internal/modules/esm/translators.js
```diff
@@ -94,3 +94,24 @@ translators.set('builtin', async function(url) {
...
+
+// Strategy for loading a wasm module
+translators.set('wasm', async function(url) {
+  const pathname = fileURLToPath(url);
+  const buffer = await readFileAsync(pathname);
+  debug(`Translating WASMModule ${url}`);
+  let result, keys;
+  try {
+    WebAssembly.validate(buffer);
+    result = await WebAssembly.instantiate(buffer, {});
```
devsnek: this doesn't support imports?

MylesBorins: just a first pass... I'm going to refer to your implementation to get support for imports

lib/internal/modules/esm/translators.js
```diff
@@ -94,3 +94,24 @@ translators.set('builtin', async function(url) {
...
 });
+
+// Strategy for loading a wasm module
+translators.set('wasm', async function(url) {
+  const pathname = fileURLToPath(url);
+  const buffer = await readFileAsync(pathname);
+  debug(`Translating WASMModule ${url}`);
+  let result, keys;
+  try {
+    WebAssembly.validate(buffer);
```
devsnek: i think this call is pointless because it wouldn't be possible to instantiate an invalid buffer anyway

MylesBorins:
@devsnek I had a meeting with some of the folks working on the WASM team today. I don't think it is clear how much of this will be in ecma262 / V8 vs. blink

Since this is related to the loader there way not actually be as much as expected inside of V8 proper. I also think it might be worth exploring this as an experimental interface to get an implementation of the expected behavior into the hands of developers, this in turn could affect the standard being developed.

i'll let the folks I cc'd chime in about timing and implementation thoughts

lib/internal/modules/esm/translators.js
MylesBorins: I was thinking it gave us a free better error 
¯\\\_(ツ)\_/¯

devsnek: @MylesBorins it doesn't throw, it returns true/false

guybedford:
@MylesBorins I think named exports are important to include here, as that is how the WASM spec integration with ES modules will work (//cc @littledan).

If you are looking for the code to do identifier validation, I included this in https://github.com/nodejs/ecmascript-modules/pull/31/commits/9377b2775e7c8987c3cd5d92e8b7fed6511ba1dd#diff-e866c1b8fc8510c620d56a7c4b841885R133 for the CommonJS named exports PR. I think that would be a worthy addition and important inclusion for merging this.

lib/internal/modules/esm/translators.js
```diff
@@ -94,3 +94,24 @@ translators.set('builtin', async function(url) {
...
+    result = await WebAssembly.instantiate(buffer, {});
+    keys = Object.keys(result.instance.exports);
+  } catch (err) {
+    err.message = pathname + ': ' + err.message;
+    throw err;
+  }
+  return createDynamicModule([...keys, 'default'], url, (reflect) => {
+    for (const key of keys)
+      reflect.exports[key].set(result.instance.exports[key]);
+    reflect.exports.default.set(result.instance.exports);
```
guybedford: This seems like the invalid line.

lib/internal/modules/esm/translators.js
```diff
@@ -94,3 +94,24 @@ translators.set('builtin', async function(url) {
...
+    WebAssembly.validate(buffer);
+    result = await WebAssembly.instantiate(buffer, {});
+    keys = Object.keys(result.instance.exports);
+  } catch (err) {
+    err.message = pathname + ': ' + err.message;
+    throw err;
+  }
+  return createDynamicModule([...keys, 'default'], url, (reflect) => {
+    for (const key of keys)
+      reflect.exports[key].set(result.instance.exports[key]);
```
guybedford: These ones may need the identifier validation? Although maybe we can actually get away without it... would be worth running some tests.

littledan:
Hey, I'm working on standards-track Wasm/ESM integration at https://github.com/WebAssembly/esm-integration/tree/master/proposals/esm-integration . I hope we can work out semantics here that match the proposed standard. I don't have any planned changes to semantics, but over the next few days I'll be trying to clean up the explainer and spec text to be more clear and well-layered with other changes.

lib/internal/modules/esm/translators.js
guybedford: I've removed this in the handling of the separation of the compile / instantiate phases for imports support.

lib/internal/modules/esm/translators.js
guybedford: Imports support now added.

lib/internal/modules/esm/translators.js
```diff
@@ -141,3 +144,32 @@ translators.set('json', async function jsonStrategy(url) {
...
+    // deadlock in cycles, both of which will be resolved when
+    // top-level await support lands.
+    const loader = await esmLoader.loaderPromise;
+    const imports = Object.create(null);
+    await Promise.all(
+      WebAssembly.Module.imports(compiled).map(async (impt) => {
+        imports[impt.module] = await loader.import(impt.module, url);
+      })
+    );
+    const { exports } = await WebAssembly.instantiate(compiled, imports);
```
devsnek: this will invoke the wasm start function too early, we need to approach this another way.

ljharb:
What are the (perhaps separate from this PR; happy to be linked elsewhere) plans to load wasm in cjs?

lib/internal/modules/esm/translators.js
```diff
@@ -141,3 +144,32 @@ translators.set('json', async function jsonStrategy(url) {
...
+// Strategy for loading a wasm module
+translators.set('wasm', async function(url) {
+  const pathname = fileURLToPath(url);
+  const buffer = await readFileAsync(pathname);
+  debug(`Translating WASMModule ${url}`);
+  try {
+    const compiled = await WebAssembly.compile(buffer);
+    // Note: This approach executes dependencies early, and will
+    // deadlock in cycles, both of which will be resolved when
+    // top-level await support lands.
```
littledan: Neither of those have to do with top-level await. Top-level await is all about avoiding calling `instantiate` from the main thread. These issues should be fixable without fixing that other problem.

lib/internal/modules/esm/translators.js
littledan: Is there any place I can find documentation for how the translators work?

devsnek: they're called during instantiation to turn absolute resolved URLs into V8 modules. running the start function at the correct time means it needs to go into that callback just below where all the exports are assigned. that callback is called during the evaluation phase.

devsnek: we should use `WebAssembly.Module.exports()` to get the export names, and then `new WebAssembly.Instance()` inside the callback.

guybedford:
My concern with `new WebAssembly.Instance` was I was under the impression that would impose the WASM 4KB limit like in browsers. But if that is not the case, then certainly we can implement the exact semantics with this approach.

guybedford:
Removing the agenda label for now, as I won't have time to implement the new semantics in time for this weeks meeting, unless someone else wants to.

devsnek:
@guybedford that limit is imposed by chrome itself, the normal api is unlimited.

guybedford:
@devsnek thanks for clarifying. Then we should be able to support this with exact semantics certainly. Feel free to work on it further too if you like. Otherwise I will get to it when I can.

lib/internal/modules/esm/translators.js
guybedford: Agreed, although this will be a larger refactoring of the module system in Node.js yes.

lib/internal/modules/esm/translators.js
```diff
@@ -162,11 +162,13 @@ translators.set('wasm', async function(url) {
         imports[impt.module] = await loader.import(impt.module, url);
       })
```
littledan: I guess this is where the deadlock in cycles comes from. It would be better to wait for the dynamic callback to actually run this. There should be a runtime TDZ on a cycle-closing edge, rather than a deadlock

devsnek: maybe we can report them up back to modulejob and have them integrated with the existing infra.

guybedford:
I've pushed up a commit to support the exact import semantics here now. Because the v8 module system can only link Module objects that are SourceTextModule objects, we need to create one to handle this, just like we do for dynamic modules.

Should be good to go now, and thanks all for review.

lib/internal/modules/esm/translators.js
guybedford: I've added this support in the latest commit.

lib/internal/modules/esm/create_wasm_module.js
```diff
@@ -0,0 +1,47 @@
...
+    );
+    const impts =
+        [...importNames].map((name) => `${name} as $${name}`).join(', ');
+    const defs = [...importNames].map((name) => `${name}: $${name}`).join(', ');
+    return `import { ${impts} } from ${specifierStr};
+    imports[${specifierStr}] = { ${defs} };`;
+  }).join('\n')}
+const { exports } = new WebAssembly.Instance(import.meta.module, imports);
+${WebAssembly.Module.exports(module)
+  .map(({ name }) => `export const ${name} = exports.${name};`).join('\n')}
```
weswigham: A facade like this means that there won't be live bindings for the wasm module, no? Are wasm exports supposed to be live bound?

guybedford:
Yes Web Assembly modules dont have live bindings, instead features like
mutable global are captured through a static binding interface.

lib/internal/modules/esm/create_wasm_module.js
```diff
@@ -0,0 +1,47 @@
...
+    const specifierStr = JSON.stringify(module);
+    const importNames = new Set(
+      imports
+      .filter(({ module: m }) => m === module)
+      .map(({ name }) => name)
+    );
+    const impts =
+        [...importNames].map((name) => `${name} as $${name}`).join(', ');
+    const defs = [...importNames].map((name) => `${name}: $${name}`).join(', ');
+    return `import { ${impts} } from ${specifierStr};
```
devsnek: i'd rather we combine this with the existing dynamic modules

ljharb:
That can’t provide live bindings, or they can’t consume them?

guybedford:
@ljharb Web Assembly modules neither import or export live bindings.

lib/internal/modules/esm/create_wasm_module.js
guybedford: Interesting. Will put some thought to it.

littledan:
What @guybedford says about WebAssembly modules matches the proposal which you can find in more detail at https://github.com/webassembly/esm-integration .

lib/internal/modules/esm/create_wasm_module.js
guybedford: Ok I've unified wasm modules with dynamic modules :)

lib/internal/modules/esm/default_resolve.js
```diff
@@ -45,7 +45,6 @@ const legacyExtensionFormatMap = {
 };
 
 if (experimentalWasmModules) {
-  // This is a total hack
   Object.assign(extensionFormatMap, {
```
devsnek: why not `extensionFormatMap['.wasm'] = 'wasm';`?

guybedford: kk

guybedford:
Upstream PR created at https://github.com/nodejs/node/pull/27659.

PR #47, opened 2019-03-04
Reinclude `--loader` temporarily, moving new work to Phase3

guybedford:
Previously, `--loader` was reverted to begin work on a new loader API. This work has progressed really well but it is clear at this point that it will need to be released as part of Phase 3.

Instead of shipping with no loaders at all, bringing back the old API while being explicit that it is being redesigned and will be replaced ensures that we can still allow userland experimentation. For example, it would then be possible to implement our final default loader API as a --loader of Node.js 12.0.0.

I've upgraded the loader implementation to follow the naming conventions of our current implementation (`"commonjs"`, `"module"` as the type strings).

### DIFF
--- a/lib/internal/process/esm_loader.js
+++ b/lib/internal/process/esm_loader.js
@@ -14,6 +14,7 @@ if (type && type !== 'commonjs' && type !== 'module')
 exports.typeFlag = type;
 
 const { Loader } = require('internal/modules/esm/loader');
+const { pathToFileURL } = require('internal/url');
 const {
   wrapToModuleMap,
 } = require('internal/vm/source_text_module');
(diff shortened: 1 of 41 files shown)


lib/internal/process/esm_loader.js
```diff
@@ -44,8 +45,15 @@ exports.loaderPromise = new Promise((resolve) => loaderResolve = resolve);
 exports.ESMLoader = undefined;
 
 exports.initializeLoader = function(cwd, userLoader) {
-  const ESMLoader = new Loader();
+  let ESMLoader = new Loader();
   const loaderPromise = (async () => {
+    if (userLoader) {
+      const hooks = await ESMLoader.import(
```
MylesBorins: Could we include a one time warning here about the loader implementation being unstable and likely to be replaced?

MylesBorins:
CI: https://ci.nodejs.org/job/node-test-pull-request/21195/

lib/internal/process/esm_loader.js
MylesBorins: an example of something similar from the `fs.promises` api

https://github.com/nodejs/node/blob/master/lib/fs.js#L1926-L1937

GeoffreyBooth:
This seems fine to me, do you mind please explaining any caveats that users should be aware of if they try to use this implementation? Like what would you say in the docs to any users considering using this?

MylesBorins:
We should also docs deprecate the prior loader

lib/internal/process/esm_loader.js
guybedford: Good point, I've added this warning.

guybedford:
Updated CI: https://ci.nodejs.org/job/node-test-pull-request/21214/

MylesBorins:
Rebased against master and pushed a fixup for the broken tests

CI: https://ci.nodejs.org/job/node-test-pull-request/21245/

MylesBorins:
48 hours, landing

Issue #16, opened 2018-11-29
Documentation: Migrating from --𝚎𝚡𝚙𝚎𝚛𝚒𝚖𝚎𝚗𝚝𝚊𝚕-𝚖𝚘𝚍𝚞𝚕𝚎𝚜 — Discrepancies and Considerations


SMotaal:
This issue is intended to track a list of aspects destined to live in a "Migrating from --𝚎𝚡𝚙𝚎𝚛𝚒𝚖𝚎𝚗𝚝𝚊𝚕-𝚖𝚘𝚍𝚞𝚕𝚎𝚜 — Discrepancies and Considerations" document. Those are a shortlist of features which users relied on previously to address certain problems for which the new implementation either offers a more robust alternative or otherwise.

- [ ] Default handling of `index.mjs` inside and outside of <kbd>ESM Package Scope</kbd>
   - TBD

Trott:
@nodejs/modules-active-members Should this issue be moved to the main node repo or closed?

MylesBorins:
I think it can be closed

PR #51, opened 2019-03-06
Docs, of the implementation as it is right this moment

GeoffreyBooth:
This PR updates the docs for where the implementation is currently. The idea is to reflect it as it is, not as it’s eventually going to be; when the expected future changes land, the docs should be updated accordingly. The old `--experimental-modules` docs are [here](https://nodejs.org/api/esm.html).

Readable versions:
- https://github.com/nodejs/ecmascript-modules/blob/docs/doc/api/esm.md
- https://github.com/nodejs/ecmascript-modules/blob/docs/doc/api/cli.md

### DIFF
--- a/doc/api/cli.md
+++ b/doc/api/cli.md
@@ -513,13 +513,13 @@ Track heap object allocations for heap snapshots.
 
 ### `-m`, `--type=type`
 
-With `--experimental-modules`, this tells Node.js whether to interpret the
+Used with `--experimental-modules`, this configures Node.js to interpret the
 initial entry point as CommonJS or as an ES module.
 
-Valid values are `"commonjs"` and `"module"`, where the default is to infer from
+Valid values are `"commonjs"` and `"module"`. The default is to infer from
 the file extension and the `"type"` field in the nearest parent `package.json`.
 
-Works with standard file input as well as `--eval`, `--print`, `STDIN`.
+Works for executing a file as well as `--eval`, `--print`, `STDIN`.
 
 `-m` is an alias for `--type=module`.
 
--- a/doc/api/cli.md
+++ b/doc/api/cli.md
@@ -513,13 +513,13 @@ Track heap object allocations for heap snapshots.
 
 ### `-m`, `--type=type`
 
-When using `--experimental-modules`, this informs the module resolution type
-to interpret the top-level entry into Node.js.
+With `--experimental-modules`, this tells Node.js whether to interpret the
+initial entry point as CommonJS or as an ES module.
 
-Works with stdin, `--eval`, `--print` as well as standard execution.
+Valid values are `"commonjs"` and `"module"`, where the default is to infer from
+the file extension and the `"type"` field in the nearest parent `package.json`.
 
-Valid values are `"commonjs"` and `"module"`, where the default is to infer
-from the file extension and package type boundary.
+Works with standard file input as well as `--eval`, `--print`, `STDIN`.
 
 `-m` is an alias for `--type=module`.
 
--- a/doc/api/esm.md
+++ b/doc/api/esm.md
@@ -49,7 +49,7 @@ For completeness there is also `--type=commonjs`, for explicitly running a `.js`
 file as CommonJS. This is the default behavior if `--type` or `-m` is
 unspecified.
 
-The `--type=module` or `-m` flags can also be used to tell Node.js to treat as
+The `--type=module` or `-m` flags can also be used to configure Node.js to treat as
 an ES module input sent in via `--eval` or `--print` (or `-e` or `-p`) or piped
 to Node.js via `STDIN`.
 
@@ -87,7 +87,7 @@ node --experimental-modules my-app.js # Runs as ES module
 If the nearest parent `package.json` lacks a `"type"` field, or contains
 `"type": "commonjs"`, extensionless and `.js` files are treated as CommonJS.
 If the volume root is reached and no `package.json` is found,
-Node.js behaves the same as if it had found a `package.json` with no `"type"`
+Node.js defers to the deafult, a `package.json` with no `"type"`
 field.
 
 ## Package Scope and File Extensions
--- a/doc/api/esm.md
+++ b/doc/api/esm.md
@@ -8,25 +8,117 @@
 <!--name=esm-->
 
 Node.js contains support for ES Modules based upon the
-[Node.js EP for ES Modules][] and the [ESM Minimal Kernel][].
+[Node.js EP for ES Modules][] and the [ecmascript-modules implementation][].
 
-The minimal feature set is designed to be compatible with all potential
-future implementations. Expect major changes in the implementation including
-interoperability support, specifier resolution, and default behavior.
+Expect major changes in the implementation including interoperability support,
+specifier resolution, and default behavior.
 
 ## Enabling
 
 <!-- type=misc -->
 
 The `--experimental-modules` flag can be used to enable features for loading
-ESM modules.
+ECMAScript modules.
 
-Once this has been set, files ending with `.mjs` will be able to be loaded
-as ES Modules.
+Once this has been set, there are a few different ways to run a file as an ES module:
+
+### <code>.mjs</code> extension
+
+Files ending with `.mjs` will be loaded as ES modules.
 
 ```sh
 node --experimental-modules my-app.mjs
 ```
+
+### <code>--type=module</code> / <code>-m</code> flag
+
+Files ending with `.js` or `.mjs`, or lacking any extension,
+will be loaded as ES modules when the `--type=module` flag is set.
+This flag also has a shorthand alias `-m`.
+
+```sh
+node --experimental-modules --type=module my-app.js
+# or
+node --experimental-modules -m my-app.js
+```
+
+For completeness there is also `--type=commonjs`, for explicitly running a
+`.js` file as CommonJS.
+This is the default behavior if `--type` or `-m` is unspecified.
+
+The `--type=module` or `-m` flags can also be used to tell Node.js to treat as an ES module input sent in via `--eval` or `--print` (or `-e` or `-p`) or piped to Node.js via `STDIN`.
+
+```sh
+node --experimental-modules --type=module --eval \
+  "import { sep } from 'path'; console.log(sep);"
+
+coffee --print file-containing-import-statements.coffee | \
+  node --experimental-modules --type=module
+```
+
+### <code>package.json</code> <code>"type"</code> field
+
+Files ending with `.js` or `.mjs`, or lacking any extension,
+will be loaded as ES modules when the nearest parent `package.json` file
+contains a top-level field `"type"` with a value of `"module"`.
+
+The nearest parent `package.json` is defined as the first `package.json` found
+when searching in the current folder, that folder’s parent, and so on up
+until the root of the volume is reached.
+
+```js
+// package.json
+{
+  "type": "module"
+}
+```
+
+```sh
+# In same folder as above package.json
+node --experimental-modules my-app.js # Runs as ES module
+```
+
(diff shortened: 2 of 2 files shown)


GeoffreyBooth:
I’ve made some updates per today’s call. Still to do:

- [ ] Section very briefly introducing `import` and `export`, linking to MDN; @evanplaice will write
- [ ] Section listing differences between ESM in Node vs browsers
- [ ] Update Specifiers section with info about Windows specifiers; @jdalton do you mind tackling?

In separate PRs:

- [ ] Blog post announcing new `--experimental-modules` (I guess this should be a PR against the modules repo)
- [ ] Text explaining migration from old `--experimental-modules` to new (another blog post?); #16

GeoffreyBooth:
I’m working on a different PR that will need its own docs, which means it needs to build off of this one. So I’d rather merge this in as is and continue to improve it in follow up PRs. The to-do items in my last comment still apply.

Any notes about what’s in this PR?

doc/api/cli.md
```diff
@@ -513,13 +513,13 @@ Track heap object allocations for heap snapshots.
 
 ### `-m`, `--type=type`
 
-When using `--experimental-modules`, this informs the module resolution type
-to interpret the top-level entry into Node.js.
+With `--experimental-modules`, this tells Node.js whether to interpret the
```
MylesBorins:
```suggestion
Used with `--experimental-modules`, this configures Node.js to interpret the
```

doc/api/cli.md
```diff
@@ -513,13 +513,13 @@ Track heap object allocations for heap snapshots.
...
+initial entry point as CommonJS or as an ES module.
 
-Works with stdin, `--eval`, `--print` as well as standard execution.
+Valid values are `"commonjs"` and `"module"`, where the default is to infer from
```
MylesBorins:
```suggestion
Valid values are `"commonjs"` and `"module"`. The default is to infer from
```

doc/api/cli.md
```diff
@@ -513,13 +513,13 @@ Track heap object allocations for heap snapshots.
...
+the file extension and the `"type"` field in the nearest parent `package.json`.
 
-Valid values are `"commonjs"` and `"module"`, where the default is to infer
-from the file extension and package type boundary.
+Works with standard file input as well as `--eval`, `--print`, `STDIN`.
```
MylesBorins:
```suggestion
Works for executing a file as well as `--eval`, `--print`, `STDIN`.
```

MylesBorins:
@GeoffreyBooth reviewing rn

doc/api/esm.md
```diff
@@ -8,37 +8,219 @@
...
+will be loaded as ES modules when the `--type=module` flag is set.
+This flag also has a shorthand alias `-m`.
+
+```sh
+node --experimental-modules --type=module my-app.js
+# or
+node --experimental-modules -m my-app.js
+```
+
+For completeness there is also `--type=commonjs`, for explicitly running a `.js`
```
MylesBorins: should this perhaps be it's own title similar to the above one?

doc/api/esm.md
```diff
@@ -8,37 +8,219 @@
...
+file as CommonJS. This is the default behavior if `--type` or `-m` is
+unspecified.
+
+The `--type=module` or `-m` flags can also be used to tell Node.js to treat as
```
MylesBorins:
```suggestion
The `--type=module` or `-m` flags can also be used to configure Node.js to treat as
```

In general I think it is better to talk about configuring rather than telling. It is a bit more accurate imho

doc/api/esm.md
```diff
@@ -8,37 +8,219 @@
...
+```
+
+### <code>package.json</code> <code>"type"</code> field
+
+Files ending with `.js` or `.mjs`, or lacking any extension,
+will be loaded as ES modules when the nearest parent `package.json` file
+contains a top-level field `"type"` with a value of `"module"`.
+
+The nearest parent `package.json` is defined as the first `package.json` found
+when searching in the current folder, that folder’s parent, and so on up
```
MylesBorins:
```suggestion
by searching recursively from the current folder until the root of the volume is reached.
```

doc/api/esm.md
```diff
@@ -8,37 +8,219 @@
...
+until the root of the volume is reached.
```
MylesBorins:
```suggestion
```

doc/api/esm.md
```diff
@@ -8,37 +8,219 @@
...
+
+```sh
+# In same folder as above package.json
+node --experimental-modules my-app.js # Runs as ES module
+```
+
+If the nearest parent `package.json` lacks a `"type"` field, or contains
+`"type": "commonjs"`, extensionless and `.js` files are treated as CommonJS.
+If the volume root is reached and no `package.json` is found,
+Node.js behaves the same as if it had found a `package.json` with no `"type"`
```
MylesBorins:
```suggestion
Node.js defers to the deafult, a `package.json` with no `"type"`
```

doc/api/esm.md
```diff
@@ -8,37 +8,219 @@
...
+field.
+
+## Package Scope and File Extensions
+
```
MylesBorins: I found the copy here a bit confusing... I'm also struggling to come up with something better.

`A package scope is defined as all children of a folder with a package.json that do not themselves have an additional package.json`

I think the above is equally confusing though sigh...

doc/api/esm.md
```diff
@@ -8,37 +8,219 @@
...
+scopes. A `package.json` lacking a `"type"` field is treated as if it contained
+`"type": "commonjs"`.
+
+The package scope applies not only to initial entry points (`node
+--experimental-modules my-app.js`) but also to files referenced by `import`
+statements and `import()` expressions.
 
-### Supported
+```js
+// my-app.js, in an ES module package scope because there is a package.json
```
MylesBorins: it might be good to have a quick ascii diagram of the filesystem

doc/api/esm.md
```diff
@@ -8,37 +8,219 @@
...
+  `"module"` package scope).
+
+- Within a `"type": "commonjs"` package scope, Node.js can be instructed to
+  interpret a particular file as an ES module by naming it with an `.mjs`
+  extension (since both `.js` and `.cjs` files are treated as CommonJS within a
+  `"commonjs"` package scope).
+
+## Package Entry Points
+
+When a `package.json` contains `"type": "module"`, its `"main"` field defines
```
MylesBorins: if you have a .cjs file in the main it will load as CJS afaik

doc/api/esm.md
GeoffreyBooth: We discussed that, and we were leaning toward no because the overall section is “how to start an ES module program”. Note that `--type` gets its own more straightforward documentation in `cli.md`.

doc/api/esm.md
GeoffreyBooth: In my experience, recursive file system searching always searches _down,_ from the current folder down into all subfolders below. This goes up.

doc/api/esm.md
GeoffreyBooth: It’s not just “all children,” it’s also the folder that the `package.json` itself is in.

doc/api/esm.md
GeoffreyBooth: Like this? https://github.com/GeoffreyBooth/node-import-file-specifier-resolution-proposal/#example

cc @SMotaal

doc/api/esm.md
ljharb: Tools that look for config always search recursively upwards - I don't think a direction is implied by "recursive" as much as by "what you're searching for" - so maybe adding `upwards` here would clarify?

doc/api/esm.md
GeoffreyBooth: Yes, this surprised me in your searching PR. I thought otherwise when I wrote this. Will update.

doc/api/esm.md
GeoffreyBooth: I’m thinking of shell commands like `cp -r`. I think users are probably more familiar with recursive meaning “here and all below” than upwards.

Babel describes how it searches for `.babelrc` [this way](https://babeljs.io/docs/en/config-files#file-relative-configuration):

> Babel loads `.babelrc` (and `.babelrc.js` / `package.json#babel`) files by searching up the directory structure starting from the “filename” being compiled

ljharb: given that package.json, eslintrc, babelrc, nvmrc, and most other rc files recursively search upwards, i don't think it's productive to argue about our differing subjective opinions about what users are more familiar with. What's objective is that "recursive searching" does *not* imply a direction.

Would the simple change of "recursively" to "recursively upwards" in myles' suggestion suffice?

GeoffreyBooth: I don’t think the word “recursive,” whether up or down, makes this any easier to understand; and for some segment of the audience it will make it more confusing. I think for that reason we should go with simpler language.

GeoffreyBooth:
Merging to unblock other PRs, but that doesn’t mean I’m rejecting the above comments; they’ll just need to be addressed in follow-on PRs. Expanded to-do list:

- [ ] Section very briefly introducing `import` and `export`, linking to MDN; @evanplaice will write
- [ ] Section listing differences between ESM in Node vs browsers
- [ ] Update Specifiers section with info about Windows specifiers; @jdalton do you mind tackling?
- [ ] Add ASCII diagram like [this](https://github.com/GeoffreyBooth/node-import-file-specifier-resolution-proposal/#example) to illustrate package scope and resolution stuff; @SMotaal ?

Maybe:

- [ ] Clearer language explaining the search for `package.json`.
- [ ] Clearer language defining “package scope.”

In separate PRs:

- [ ] Blog post announcing new `--experimental-modules` (I guess this should be a PR against the modules repo)
- [ ] Text explaining migration from old `--experimental-modules` to new (another blog post?); #16

doc/api/esm.md
```diff
@@ -289,128 +659,11 @@ READ_PACKAGE_JSON(_packageURL_)
...
-      // Get and set functions provided for pre-allocated export names
-      exports.customExportName.set('value');
-    }
-  };
-}
-```
-
-With the list of module exports provided upfront, the `execute` function will
-then be called at the exact point of module evaluation order for that module
-in the import tree.
```
guybedford: Was this section intended to be removed?

doc/api/esm.md
```diff
@@ -8,37 +8,223 @@
...
+
+## Package Scope and File Extensions
+
+A folder containing a `package.json` file, and all subfolders below that
+folder down until the next folder containing another `package.json`, is
+considered a _package scope_. The `"type"` field defines how `.js` and
+extensionless files should be treated within a particular `package.json` file’s
+package scope. Every package in a project’s `node_modules` folder contains its
+own `package.json` file, so each project’s dependencies have their own package
+scopes. A `package.json` lacking a `"type"` field is treated as if it contained
```
guybedford: This isn't quite true - because we check the extension and throw on conflicts when using "type", while when there is no type there are no conflict errors.

doc/api/esm.md
GeoffreyBooth: Crap. That must’ve gotten lost in a rebase.

GeoffreyBooth: Wait, no, it was just moved up. The idea was to have the spec be last.

doc/api/esm.md
GeoffreyBooth: We only throw on conflicts for initial entry points. I was referring to `import` statements of files within scopes here, though I acknowledge that I didn’t make that explicit 😄.

Do we need to document the cases for which we error?

GeoffreyBooth: I rewrote the flags sections as part of https://github.com/nodejs/ecmascript-modules/blob/525e8c9b5d0acc7b8b7895f7ea65b592fdf3e2bd/doc/api/cli.md#--typetype--m--a

PR #52, opened 2019-03-07
doc: improve createRequire example

MylesBorins:
Original doesn't really make sense in retrospect

updates example to

```mjs
// cjs.js
module.exports = 'cjs';

// esm.mjs
import { createRequireFromPath as createRequire } from 'module';
import { fileURLToPath as fromURL } from 'url';

const require = createRequire(fromURL(import.meta.url));

const cjs = require('./cjs');
cjs === 'cjs'; // true
```

### DIFF
--- a/doc/api/esm.md
+++ b/doc/api/esm.md
@@ -341,9 +341,9 @@ module.exports = 'cjs';
 
 // esm.mjs
 import { createRequireFromPath as createRequire } from 'module';
-import { fileURLToPath as fromPath } from 'url';
+import { fileURLToPath as fromURL } from 'url';
 
-const require = createRequire(fromPath(import.meta.url));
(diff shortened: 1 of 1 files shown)


GeoffreyBooth:
Can you please target the `docs` branch?

MylesBorins:
@GeoffreyBooth done

GeoffreyBooth:
Out of scope for this PR, but I just read the [`createRequireFromPath` docs](https://nodejs.org/api/modules.html#modules_module_createrequirefrompath_filename) and they don’t explain _why_ you’d want a `require` function where the argument it’s given is relative to some folder other than the one the current file is in. It seems pretty pointless outside of the ESM `import.meta.url` context.

PR #53, opened 2019-03-08
Modules lkgr rebase

MylesBorins:
another day another conflict.


MylesBorins:
Took the liberty to squash some of the duplicated commits related to the loader revert.

CI: https://ci.nodejs.org/job/node-test-pull-request/21347/

MylesBorins:
oh fun things are actually broken... going to dig in

guybedford:
Fun indeed. Looks like it's - https://github.com/nodejs/node/commit/b05fd4baa87886674721101eaf38b75716037891, where the ESM init should happen after debug log.

guybedford:
I've only glanced at it briefly, but something like moving this require - https://github.com/nodejs/node/blob/master/lib/internal/main/check_syntax.js#L14 so that it is inside of the function body after prepareMainThreadExecution() may help.

MylesBorins:
Took liberty to clean up the branch for upstreaming. 35 commits down to 7 (+1 for test fix). Did my best to split it up into observable bits. Still will need to go through it again and ensure that the test suite passes for each commit

@guybedford @GeoffreyBooth @jdalton you all had commits that got squashed at some point. I've included all authors in commit meta data though.

CI: https://ci.nodejs.org/job/node-test-pull-request/21364/

GeoffreyBooth:
Looks like the CI build failed?

MylesBorins:
@GeoffreyBooth looks like my solution introduced a "double require" which the linter is yelling about. Will need to refactor a bit, but going to compare against upstream first

MylesBorins:
another try at CI: https://ci.nodejs.org/job/node-test-pull-request/21386/

lib/internal/main/check_syntax.js
```diff
@@ -13,12 +13,16 @@ const {
...
 
-const CJSModule = require('internal/modules/cjs/loader');
 const vm = require('vm');
 const {
   stripShebang, stripBOM
 } = require('internal/modules/cjs/helpers');
 
+let CJSModule;
+function CJSModuleInit() {
+  if (!CJSModule)
```
MylesBorins: this guard is probably not necessary... but I'm being paranoid

MylesBorins:
Did another pass on the rebase. Simplified commits and ensured the test suite passes for each one

I think we could fold 89ddcfd, a9887b8, and c7e5fe0 into one commit potentially... for the IRP implementation. Haven't dug into if that diff is easier to review. Thoughts?

CI: https://ci.nodejs.org/job/node-test-pull-request/21410/

GeoffreyBooth:
> I think we could fold [89ddcfd](https://github.com/nodejs/ecmascript-modules/commit/89ddcfd6bd7577fbfbbd93d546ccafd7577b4fcf), [a9887b8](https://github.com/nodejs/ecmascript-modules/commit/a9887b8aad3eda125340344d44a4a7f74a9945d6), and [c7e5fe0](https://github.com/nodejs/ecmascript-modules/commit/c7e5fe088893cf0051265b42d1154c68645e091b) into one commit potentially… for the IRP implementation. Haven’t dug into if that diff is easier to review. Thoughts?

Those three commits are basically: our Phase 0 work (remove stuff we’ll be replacing); add IRP; revert removal of loader code. I could see maybe the first and third being combined, but adding the IRP implementation feels like it should be its own commit.

MylesBorins:
@GeoffreyBooth my thought process was that upstream there isn't necessarily any reason to keep this more granual, it is an artifact of our process. The loader revert is someone dependent on the types implementation in IRP... so it is a non-trivial amount of extra work to keep the tree clean.

Keep in mind there is a non-zero chance all our work will be squashed into a single commit when landing on core. This is more about making review easier

GeoffreyBooth:
Sure, whatever you think is best.

MylesBorins:
landed in 2e2670a207...ea592210c6

PR #58, opened 2019-03-10
Cpp Refactoring Ensuring avoiding duplicate package cache

guybedford:
This brings through the refactorings from #57 that avoid cache duplication and should improve performance, described in detail there, while retaining the current behaviour of having the `--type` flag only apply to the entry point file and not any other files in the entry point scope.


MylesBorins:
landed in c894000fae

PR #61, opened 2019-03-14
esm: fix extensionless import for --type

MylesBorins:
At some point this regressed for extensionless imports
for esm.

### DIFF
--- a/lib/internal/modules/esm/default_resolve.js
+++ b/lib/internal/modules/esm/default_resolve.js
@@ -4,6 +4,7 @@ const { NativeModule } = require('internal/bootstrap/loaders');
 const { ERR_TYPE_MISMATCH,
         ERR_UNKNOWN_FILE_EXTENSION } = require('internal/errors').codes;
 const { getOptionValue } = require('internal/options');
+const optionType = getOptionValue('--type');
 const experimentalJsonModules = getOptionValue('--experimental-json-modules');
 const { resolve: moduleWrapResolve } = internalBinding('module_wrap');
 const { pathToFileURL, fileURLToPath } = require('internal/url');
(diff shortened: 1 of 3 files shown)


MylesBorins:
CI: https://ci.nodejs.org/job/node-test-pull-request/21552/

MylesBorins:
@GeoffreyBooth it currently only looks at the type field from the package.json, which in this implementation would trump the type flag (perhaps not what we want?)

lib/internal/modules/esm/default_resolve.js
```diff
@@ -75,15 +76,15 @@ function resolve(specifier, parentURL) {
       }
```
guybedford: I think the simpler fix would just be a single line here on line 77:

```suggestion
      }
      if (!format) {
        format = asyncESM.typeFlag;
      }
```

Note also that the `type === 0` check you have done is effectively saying we shouldn't be able to override when there is an existing "type" in the package.json. If that is what we want, we could explicitly change the `if (ext === '.js')` check in the validations above to `if (ext !== '.cjs' && ext !== '.mjs')` so that we get a proper validation error in this case.

guybedford:
Yes extensionless support for `package.json` "type" sounds like a separate bug. Perhaps we can tackle that separately.

lib/internal/modules/esm/default_resolve.js
```diff
@@ -75,15 +76,15 @@ function resolve(specifier, parentURL) {
...
     }
   }
-
   if (!format) {
-    if (isMain)
+    if (isMain && type === 0 && optionType)
+      format = optionType;
+    else if (isMain)
       format = type === 2 ? 'module' : 'commonjs';
```
guybedford: Actually, no, this path should be supporting the package.json type for extensionless files.

MylesBorins:
Updated,  I think it has the correct behavior right now

MylesBorins:
another CI: https://ci.nodejs.org/job/node-test-pull-request/21553/

MylesBorins:
failure is unrelated. Will land in 2 hours if there are no objections... necessary for upstreaming

PR #63, opened 2019-03-15
esm: cpp refactoring update not to use uv_fs_realpath

guybedford:
I didn't realise the `uv_fs_realpath` is explicitly something to avoid in the previous refactoring, which only came up with the CI error that the windows drive name was being capitalized.

This adjusts the previous refactoring to share the package cache without using uv_fs_realpath and sticking to the JS implementation.

@MylesBorins this should squash well into the previous refactoring commit.


MylesBorins:
rebased

CI: https://ci.nodejs.org/job/node-test-pull-request/21573/

PR #68, opened 2019-04-04
Replace --entry-type with --input-type, plus --input-type=auto

GeoffreyBooth:
#66 plus the addition of `--input-type=auto`. Same logic as #55: if the input string contains an `import` or `export` statement, treat it as ESM; otherwise as CommonJS. This is perhaps even more useful for `--eval` and `--print`.

I’m submitting this as a separate PR in case people are okay with #66 but uncomfortable with this addition. You can see the difference between the PRs here:
> https://github.com/nodejs/ecmascript-modules/compare/flag-for-string-input-only…flag-for-string-input-only-with-auto

It’s the same code as from #55, without `-a`.


ljharb:
@GeoffreyBooth if you set the base branch of this PR to #66's, then the diff will be reduced and readable.

GeoffreyBooth:
I thought that’s what the link was doing, but sure.

lib/internal/modules/esm/default_resolve.js
```diff
@@ -94,6 +94,12 @@ function resolve(specifier, parentURL) {
     // entry point, etc.).
     throw new ERR_INPUT_TYPE_NOT_ALLOWED();
   }
+  // --input-type=auto detects an ESM .js file within a CommonJS scope.
+  if (isMain && format === 'commonjs' && typeFlag === 'auto') {
```
ljharb: clarifying again, just to be sure: `isMain` here only applies to the entry point, and not to any other file that's required?

GeoffreyBooth:
> Repeating my review from #55; If we can’t detect the type unambiguously from syntax, it should throw.

I’m not sure what this means. What are you proposing?

The definition of `auto` is in the docs in this PR: run as ESM if an `import` or `export` statement is detected, else run as CommonJS. That’s the only “foolproof” detection algorithm possible. The user simply needs to know that that’s what “auto” means—it’s not affirmatively detecting either mode. It’s detecting `import` or `export` statements, period.

lib/internal/modules/esm/default_resolve.js
GeoffreyBooth: Yes, `isMain` is the initial entry point only. But `--input-type` should never apply to files at all, this is a mistake. I updated #55 too and somehow some of my changes there got applied here.

GeoffreyBooth:
Per 2019-04-10 meeting, I’ll refactor this into a separate PR that creates a new function on `module`, e.g. `import { detectImportOrExportStatement } from 'module'` or whatever it ends up being named. Then Node is providing this as an API for tools like `babel` and `coffee` to have a common, Node-supported way of picking a type to use for unknown JavaScript string input. The tools could use the result of that call in deciding whether to use `vm.runScript` or `vm.runModule` or whatever the recommended method is for them to run JavaScript code.

Once that’s done and merged in, we can discuss extending its use to other locations such as `--input-type=auto`.

ljharb:
(Let's make sure the API is usable from CJS as well 😄  )

GeoffreyBooth:
Closing in favor if https://github.com/nodejs/ecmascript-modules/pull/69

